PR #52, opened 2019-09-26
Unsubscribe all accounts from AWS spam

abstrask:
I'm tired of AWS spam in our shared mailbox.

This should be the `curl` command to opt-out of all and will be run after each account has been setup.

In this setup, it will run the command every time the module is used with a Terraform `apply` action. However, the endpoint uses to unsubscribe may not be as resilient as the AWS API we normally use, so this could theoretical make the module more fragile.

Once we're reasonably certain that we have unsubscribed from all accounts, we might consider moving the command to the "local-exec" block within the "aws_organizations_account" resource - this will cause it to only fire when creating accounts.

After merge, we should publish a new release and update `pre-prime`, `prime` and `aws-account-manifests` to use it.

### DIFF
--- a/_sub/security/org-account/main.tf
+++ b/_sub/security/org-account/main.tf
@@ -12,6 +12,8 @@ resource "aws_organizations_account" "org_account" {
 
 resource "null_resource" "ubsubscribe_spam" {
 
+  depends_on = ["aws_organizations_account.org_account"]
+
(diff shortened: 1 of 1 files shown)


runeabrahamsson:
Awesome idea. I was getting tired of the spam as well since I go look for warnings once in a while.
Good job

Issue #61, opened 2019-11-27
Traefik access log should not show requests returning HTTP 200


abstrask:
At least until we collect logs in a proper tool, we should filter out Traefik access logs to exclude 1xx and 2xx status codes.

This is done with the CLI argument `--accessLog.filters.statusCodes="300-399,400-499,500-599"`.

wcarlsen:
@abstrask didn't you solve this?

abstrask:
@wcarlsen Only in production - not in code - so it has probably been overwritten by now.

wcarlsen:
@abstrask it is running on Hellman. I'm closing this issue.

PR #71, opened 2019-12-12
Exclude 1xx and 2xx from Traefik logs, to ease spotting errors

wcarlsen:
Solves #61.


wcarlsen:
@abstrask så skulle være i kode også!

PR #76, opened 2020-01-24
Simple QA test for all EKS modules

abstrask:
The specs and scripts needed to plan, build, build again, destroy an EKS cluster, that should be *very* similar to our production cluster.

Except the test runs against the master branch of our Infrastructure Modules, so we continuously verify that it is in a deployable state.


abstrask:
@HenrikStanley Could you specifically check the pipeline spec, and ensure that it doesn't run on every commit. I imagine we only want to run this scheduled, maybe even just once a week.

PR #79, opened 2020-02-11
Removed obsolete ecr-repo, because it is unused

wcarlsen:
We do not use this in the ECRRepo. So no need for it I guess.


runeabrahamsson:
I agree we don't use the module for ECR and it is adding unnecessary maintenance for our code. Let's remove it

abstrask:
This was used to show teams a way to handle their ECR repos themselves, rather than being dependent on us creating them.

I thought it was also used for deploying Harald. At least that's what we agreed, as far as I remember.

wcarlsen:
Harald was suppose to use it, but we ran into issues I think. So Harald is only doing a postgres with Terraform https://github.com/dfds/harald/tree/master/infrastructure.

wcarlsen:
We can always reintroduce if we want to.

abstrask:
See link - it is provisioning its own ECR repo.

PR #83, opened 2020-02-18
Migrated ECR repo one to one and modularized (if that's a word) it.

wcarlsen:
A one to one migration of ECR repo module. You still have to migrate states.


HenrikStanley:
modularized adjective
mod·​u·​lar·​ized | \ ˈmä-jə-lə-ˌrīzd

Definition of modularized
1 : containing or consisting of modules
2 : produced in the form of modules

It is a word and it seems like to be used appropriately

PR #97, opened 2020-04-30
Seek kubeconfig and null_resource, and destroy! [ch2556]

abstrask:
Got rid of null_resources and removed nearly all need for "kubeconfig" file. Enough to support `plan` on `services` sub-module (still requires cluster to have been created).

Traefik module refactored and easilier convertible to v2.x.

Manually delete cluster role "traefik-ingress-controller" after deployment, as it has been replaced by "traefik-cr", and was previously deployed via a `null_resource`:

```
kubectl delete clusterrole traefik-ingress-controller
```


abstrask:
Squash and merge, but don't delete branch - more to come! :)

PR #99, opened 2020-05-29
Update db minor version from 10.9 to recommended minor version 10.11

JohannesEH:
As we're currently using the default setting on `auto_minor_version_upgrade` (which is `true`), a provisioned db could update it's minor version during the weekly maintenance window. This will break any scripts because the version is hard-coded in the terraform template. This is a short-term fix for our databases but we should probably make both `auto_minor_version_upgrade` and `engine_version` configurable. The currently recommended major postgres version on RDS is 11.x, so one could argue that that should be the default in this template as well.


JohannesEH:
This PR will be obsolete and should be closed if you accept PR #100

JohannesEH:
I close this, as it was never reviewed. And is now obsolete as the version is now at 12.

PR #100, opened 2020-05-29
Made more options configurable on the postgres module

JohannesEH:
I added `engine_version`, `auto_minor_version_upgrade` and `deletion_protection` to the list of configurable options in the private (_sub) postgres module. The most notable change is the default value `false` for `auto_minor_version_upgrade` which is normally `true`. But I think `false` is a more sane default in a Terraform/IAC workflow.

The `engine_version` can be set by two variables (db_engine_major_version & db_engine_minor_version). The major version is also used in the parameter group definition.

I added `deletion_protection` to allow users to set this value in production environments. **Note:** I set the default to `true` in the public module, as it seems we're using production defaults (environment="prod", skip_final_snapshot=false).

I also updated the public postgres module to expose all the underlying variables from the internal module.

Finally I updated the README.md file to reflect these changes and added emphasis on required and notable parts of the document.

Final notes. I feel like having our own defaults or using terraform defaults for variables such as `environment`, `skip_final_snapshot`, `deletion_protection`, `backup_retention`, `delete_automated_backups` etc. is a bad decision. As it will keep the users of this module from thinking about and/or deciding how they want to run their production databases. I don't know if it possible to provide various overrideable sets of default values (one for prod one for testing) in terraform, so we could try and help people. But personally I would really like to be able to override **all** options on the base `aws_db_instance` module so I don't have to rely on defaults if I don't want to. Another thing is that the current module is a little unsafe in regards to deletion protection, which will end up causing pain for people (it already did to me). Sure it does take a final snapshot but it will take some time getting back online if you accidentally do a destructive action. So I really think we should set `deletion_protection=true` as default, but let people consciously turn it off in non-prod environments.


JohannesEH:
**Note:** All changes in this PR _should_ be backwards compatible. But it will set `auto_minor_version_upgrade=false` and `deletion_protection=true` on existing installations.

JohannesEH:
I close this, as it was never reviewed. I still feel these changes apply to the current module though.

PR #116, opened 2020-10-02
EKS: Force update of release and recreate pods, when KIAM properties are updated

abstrask:


wcarlsen (approved):
That recreate pods will remove a lot of toil when updating KIAM

PR #120, opened 2020-10-28
Updated CNI EKS addon version mamp

wcarlsen:
Docs:
https://docs.aws.amazon.com/eks/latest/userguide/update-cluster.html

image


abstrask (commented):
@dfds-pwes and I deliberately only updated the versions for 1.16 and upwards. Are we sure we want to update it for our current version? Of course we can still just not deploy EKS pipeline with the new release, while on 1.15. Or we may specifically want to update CNI first. Thoughts?

wcarlsen:
In the case that we end up postponing the upgrade, I guess we want to keep the addon map updated.

abstrask:
That's also super cool @wcarlsen. But do we maybe specifically want to upgrade the CNI ahead of the 1.16 upgrade. In case something does go wrong, it may help that we change as few things as possible at a time? E.g. CNI now, control plane/nodes next week.

abstrask (approved):

PR #122, opened 2020-11-06
Goldpinger helm3

wcarlsen:
Goldpinger port to Terraform submodule.


abstrask (changes requested):
Super cool!

Suggest to prefix goldpinger module and variables with "monitoring_" in k8s-services, except for variable "goldpinger_namespace", which should just be "monitoring_namespace" (same as Promtheus Operator), which will be added later.

We also need to handle the "monitoring" namespace in the pipeline, otherwise we won't be able to spin-up the cluster end to end, but will have to create it manually between cluster and services phase. QA will fail if we merge this.

wcarlsen:
I guess QA wont fail for now since Goldpinger is feature toggled and disabled by default.

abstrask (approved):
LGTM.

I might consider `module "goldpinger" {...}` to module "monitoring_goldpinger" {...}, but that's just my preference. 

Nice work!

PR #129, opened 2020-11-11
Kiam qos and toleration

abstrask:
Objective:

- Classifies KIAM QoS as "Burstable" (by setting resource requests)
- Tolerate all node taints


wcarlsen (approved):
LGTM

PR #140, opened 2020-11-18
Add helm metrics-server

rifisdfds:
This PR adds a metrics-server submodule to allow our metrics-server to be managed in state/pipeline, which it currently is not


wcarlsen (approved):
Lgtm

PR #145, opened 2020-12-04
Added module for csi driver

dfds-pwes:
Implemented a module to allow the installation of the new AWS EBS CSI Driver using a Helm Chart.  Also includes the definition of associated IAM Policy, Role and annotations required for it to work with KIAM.

Note that the KIAM whiteListRouteRegexp is also updated because the CSI Driver requires additional access.


rifisdfds:
I can't see where the required annotation is being added to kube-system for the CSI Role. Is the intention that this be done here or elsewhere? Without it I don't believe the volume can be created

abstrask (approved):
Approved (with changes)

PR #147, opened 2020-12-09
removed aggregatedapidown alert silence

rifisdfds:
Removed AggregatedAPIDown alert silence (null receiver) from Prometheus. Silence is no longer required after upgrade to Kubernetes 1.18


wcarlsen (approved):
LGTM!

PR #148, opened 2020-12-15
Made test reflect we are on k8s v1.18

wcarlsen:
Removed tests for k8s 1.16 and 1.17 and included 1.19 and prepared for 1.20


abstrask (approved):
Good call, @wcarlsen!

I've changed the order of the clusters in the pipeline spec. I couldn't stand 19, 20, 18 :)

Also updated NG2 instance types to reflect prod.

PR #150, opened 2020-12-18
Set target group dereg delay to zero

abstrask:
When scalling down ASGs during maintenance, there's a 5 minute delay for each ASG, meant to allow connections to train from the ALB.

However the maintenance process itself drains the workload from the nodes before triggering a scale-in operation, so it seems pointless to wait an arbitrary 5 minutes after. This reduces the delay to 0 seconds, which should remove quite a bit of overhead from the process.

Verify in sandbox cluster.


wcarlsen (approved):
LGTM!

PR #152, opened 2021-01-06
Added locals to allow multiple roles to be assumed from kube-system

dfds-pwes:
Together with Rasmus we added some locals to the Terraform code to allow us to support multiple roles being annotated on the kube-system namespace.


wcarlsen (approved):
LGTM. Nice work!

SEQUOIIA:
Same

PR #153, opened 2021-01-06
Needed to add single quotes to kubectl command

dfds-pwes:
Needed to add single quotes to kubectl command as annotations may now include the pipe delimiter.

The absence of the quotes was causing the pipeline to fail.  This should resolve the issue.


SEQUOIIA (approved):
Taketh my approval!!

PR #162, opened 2021-02-01
Fix atlantis for flux

rifisdfds:
- add flux github token for atlantis
- fix creation of env vars to match atlantis_ prefix var name changes
- set defaults for values that end up as env vars to empty string to avoid errors if left absent (i.e if flux is disabled you do not need to set the flux github token for flux)
- remove duplicate vars in submodule to be consistent with other submodules

### DIFF
--- a/_sub/compute/helm-atlantis/main.tf
+++ b/_sub/compute/helm-atlantis/main.tf
@@ -104,6 +104,7 @@ resource "kubernetes_secret" "gh" {
 
     data = {
         github_token = var.github_token
+        github_token_flux = var.platform_fluxcd_github_token
     }
     depends_on = [ kubernetes_namespace.namespace ]
 }
--- a/compute/k8s-services/vars.tf
+++ b/compute/k8s-services/vars.tf
@@ -459,6 +459,11 @@ variable "atlantis_github_token" {
   default     = null
 }
 
+variable "atlantis_platform_fluxcd_github_token" {
+  description = "Github token that the provider uses to perform Github operations for Flux."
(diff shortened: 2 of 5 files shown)


_sub/compute/helm-atlantis/main.tf
```diff
@@ -104,6 +104,7 @@ resource "kubernetes_secret" "gh" {
 
     data = {
         github_token = var.github_token
+        github_token_flux = var.platform_fluxcd_github_token
```
abstrask, line 107: Time to merge these two providers into one, rather than leaking this unnecessary complexity into Atlantis?

compute/k8s-services/vars.tf
```diff
@@ -501,25 +506,25 @@ variable "atlantis_webhook_events" {
 
 variable "atlantis_namespace" {
   type        = string
-  description = ""
+  description = "Namespace for Atlantis deployment"
```
abstrask, line 509: 👍 for descriptions

compute/k8s-services/vars.tf
```diff
@@ -531,48 +536,48 @@ variable "atlantis_image_tag" {
 
 variable "atlantis_arm_tenant_id" {
   type        = string
-  description = ""
-  default     = null
+  description = "Used to set environment variable for ARM tenant ID"
+  default     = ""
```
abstrask, line 540: Having recently started using `null` as a default for many variables myself, what is the rational for changing from `null` to an empty string?

abstrask (approved):

PR #164, opened 2021-02-03
Qa csi debug

abstrask:
- KIAM no longer optional, and the sub module code a lot cleaner as a result
- QA refers to modules locally, enabling faster local troubleshooting
- QA EKS 1.19 and 1.20 traces removed, as those versions are not available yet 
- Destroy steps in QA pipeline skipped if variable `NO_DESTROY=true`


wcarlsen (approved):
LGTM

PR #165, opened 2021-02-09
Added Metrics server and goldpinger to QA env

wcarlsen:
Included these two services into QA environment.

### DIFF
--- a/test/integration/eu-west-1/k8s-qa18/services/terragrunt.hcl
+++ b/test/integration/eu-west-1/k8s-qa18/services/terragrunt.hcl
@@ -22,7 +22,6 @@ inputs = {
 
   eks_cluster_name = "qa18"
 
-
   # --------------------------------------------------
   # Traefik
   # --------------------------------------------------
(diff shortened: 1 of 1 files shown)


test/integration/eu-west-1/k8s-qa18/services/terragrunt.hcl
```diff
@@ -74,4 +73,21 @@ inputs = {
...
+  # Monitoring namespace
+  # --------------------------------------------------
+
+  monitoring_namespace_deploy = true
+
+  # --------------------------------------------------
+  # Metrics server
+  # --------------------------------------------------
+
+  monitoring_metrics_server = true
```
rifisdfds, line 86: Chart version missing to match production

test/integration/eu-west-1/k8s-qa18/services/terragrunt.hcl
```diff
@@ -74,4 +73,21 @@ inputs = {
 
   kiam_chart_version = "5.9.0" # 5.10.0 includes https://github.com/uswitch/kiam/pull/427, but image does not - breaks WhitelistRouteRegexp/AllowRouteRegexp
 
+  # --------------------------------------------------
+  # Monitoring namespace
+  # --------------------------------------------------
+
+  monitoring_namespace_deploy = true
```
rifisdfds, line 80: monitoring_namespace_iam_roles not set as per production

rifisdfds (changes requested):
The commented changes will need making to match production, if that is the intention of this QA

test/integration/eu-west-1/k8s-qa18/services/terragrunt.hcl
wcarlsen: Because it is not needed here for now. It defaults to empty string.

test/integration/eu-west-1/k8s-qa18/services/terragrunt.hcl
wcarlsen: Are we matching production in the QA? I thought the point was also to test upstream changes, maybe I'm wrong here. I'll just apply the chart version for now.

test/integration/eu-west-1/k8s-qa18/services/terragrunt.hcl
wcarlsen: This makes it so retarded that we set default chart versions to null then.

rifisdfds (approved):

PR #174, opened 2021-02-12
readd prometheus-stack to qa

rifisdfds:
- uncomment kube-prometheus-stack for qa
- explicitly set gp2 storageclass as per prod to prevent identified issue with deletion


dfds-pwes (approved):
Looks good to me.

PR #177, opened 2021-02-16
fixes for servicemonitor dependencies

rifisdfds:
- add depends_on to kiam module
- remove true hardcode from goldpinger

This PR adds a depends_on to the kiam module which will deploy a servicemonitor if prometheus_stack is deployed, and deploy after prometheus_stack. The depends_on is ignored if prometheus_stack is not deployed. Goldpinger now follows the same behaviour rather than always trying to deploy a servicemonitor, now it will only use one if prometheus_stack is deployed


dfds-pwes (approved):
LGTM

PR #178, opened 2021-02-16
enable kube-prometheus-stack in qa after fix

rifisdfds:
This PR enables kube-prometheus-stack in QA which should now make use of the servicemonitor dependency fix


dfds-pwes (approved):
Looks good to me.....

PR #180, opened 2021-02-16
set atlantis module to generate webhook password

rifisdfds:
This PR removes the provided password from Atlantis for the webhooks it creates and instead uses random_password to generate one. This is preferable as we do not need to see or reuse this password


dfds-pwes (approved):
Looks good to me

PR #181, opened 2021-02-17
Add atlantis to qa

rifisdfds:
This PR adds Atlantis to QA. As part of doing so, the ability to define the StorageClass has also been implemented


dfds-pwes (approved):
Looks good to me.

PR #183, opened 2021-02-23
Flux version lock

dfds-pwes:
Various changes including Flux version lock and modification of the way in which manifests are applied and synched.


abstrask (approved):
We need to include a migration note in release description re: new state addresses for manifests.

PR #185, opened 2021-02-25
Add Flux CRD delay script.

dfds-pwes:
Previously a fixed timeframe delay was used but this caused failures in QA because CRDs were not fully deployed before being utilised.  This change introduces a small bash script that will cause a delay until the CRD's are in place.

### DIFF
--- a/test/integration/eu-west-1/k8s-qa18/services/terragrunt.hcl
+++ b/test/integration/eu-west-1/k8s-qa18/services/terragrunt.hcl
@@ -56,9 +56,10 @@ inputs = {
   # Platform Flux CD
   # --------------------------------------------------
 
-  // platform_fluxcd_deploy       = true
-  // platform_fluxcd_repo_name    = "platform-manifests-qa"
-  // platform_fluxcd_github_owner = "dfds"
+  platform_fluxcd_deploy       = true
+  platform_fluxcd_release_tag  = "v0.8.2"
+  platform_fluxcd_repo_name    = "platform-manifests-qa"
+  platform_fluxcd_github_owner = "dfds"
 
   # --------------------------------------------------
   # AWS EBS CSI Driver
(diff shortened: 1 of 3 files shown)


test/integration/eu-west-1/k8s-qa18/services/terragrunt.hcl
```diff
@@ -56,9 +56,10 @@ inputs = {
   # Platform Flux CD
   # --------------------------------------------------
 
-  // platform_fluxcd_deploy       = true
-  // platform_fluxcd_repo_name    = "platform-manifests-qa"
-  // platform_fluxcd_github_owner = "dfds"
+  platform_fluxcd_deploy       = true
+  platform_fluxcd_release_tag  = "v0.8.2"
```
abstrask, line 60: Don't we generally favour 'latest', i.e. not define, for QA?

test/integration/eu-west-1/k8s-qa19/services/terragrunt.hcl
```diff
@@ -56,9 +56,10 @@ inputs = {
   # Platform Flux CD
   # --------------------------------------------------
 
-  // platform_fluxcd_deploy       = true
-  // platform_fluxcd_repo_name    = "platform-manifests-qa"
-  // platform_fluxcd_github_owner = "dfds"
+  platform_fluxcd_deploy       = true
+  platform_fluxcd_release_tag  = "v0.8.2"
```
abstrask, line 60: Don't we generally favour 'latest', i.e. not define, for QA?

abstrask (changes requested):
LGTM, but what are your thoughts on the release tag?

test/integration/eu-west-1/k8s-qa19/services/terragrunt.hcl
dfds-pwes: Yes.  Can I actually put latest there though or do I just omit the value entirely?

PR #188, opened 2021-03-08
added special char override to random_password

rifisdfds:
This PR will restrict the use of certain characters in the random_password resource that can cause issues by using override_special

For example, if `{` is the first generated character in the password it will cause the issue `Error: failed parsing key "github.secret" with value {+H8cc&y1g[%Ny2), list must terminate with '}'` so we do not want to generate this character

### DIFF
--- a/_sub/compute/helm-atlantis/main.tf
+++ b/_sub/compute/helm-atlantis/main.tf
@@ -10,6 +10,7 @@ locals {
 resource "random_password" "webhook_password" {
   length = 16
   special = true
+  override_special = "!@#$%&*-_=+:?"
 }
 
 resource "helm_release" "atlantis" {


abstrask (approved):
LGTM.

Longer term, I think we should start leaning towards generating passwords (e,g. using https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/string), rather than supplying them. We can then push them to parameter store.

wcarlsen (approved):
LGTM!

rifisdfds:
@raras-dfds That's what this is doing using [random password](https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/password) (identical to random_string except it's treated as sensitive so never displayed), the module is also set up so that you can provide a password if you really want to for any reason (sandbox etc). A parameter store is a good idea, but for a webhook like this we never actually need to know the generated password so it's best if it's not pushed anywhere at all

PR #194, opened 2021-03-30
Velero

wcarlsen:
Velero bucket storage

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,3 +48,6 @@ validate.json
 *.sum
 *.mod
 _sub/monitoring/alarm-notifier/lambda/main
+
+#Intellij
+.idea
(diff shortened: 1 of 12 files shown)


dfds-pwes (commented):
Looks good to me - but just so i'm clear this implementation is only to be used for non-production clusters right (so, for sandbox and QA mainly?).

wcarlsen:
The implementation of Velero s3 bucket main module is meant for production use also. We just need to start with QA and sandbox first.

rifisdfds (approved):

rifisdfds:
All looks fine to me. QA 1.18 is failing from it but I think it's an unrelated coincidence

PR #196, opened 2021-04-12
add crossplane metrics toggle

rifisdfds:
This PR lets you enable metrics for Crossplane using the crossplane_metrics_enabled variable set to true or false. It will default to true if crossplane is deployed


wcarlsen (approved):
LGTM, but I don't think we need the variable for disable or enable. From my point of view it could just have been hardcoded

PR #197, opened 2021-04-13
Deprecated version in provider block

dingobar:
Since terraform 0.13 this is deprecated and gives this warning:
```sh
Warning: Version constraints inside provider configuration blocks are deprecated

  on .terraform/modules/database/database/postgres/main.tf line 12, in provider "aws":
  12:   version = "~> 2.43"
```

Aws provider is also now in v3, think it would be good to upgrade at some point.

Also, removing backend to let module users decide the backend themselves.

### DIFF
--- a/database/postgres/main.tf
+++ b/database/postgres/main.tf
@@ -3,16 +3,17 @@
 # --------------------------------------------------
 
 terraform {
-  backend "s3" {
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = "~> 2.43"
+    }
   }
+  required_version = ">= 0.14"
 }
 
 provider "aws" {
   region  = var.aws_region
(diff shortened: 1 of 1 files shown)


wcarlsen:
I need your eyes on this @abstrask.

abstrask (changes requested):
It seems you have removed the "backend" configuration, which needs to be retained for main modules.

I would prefer putting the various version constraints (terraform + providers) in a separate `versions.tf` file (while retaining the `terraform { backend "s3" }` block in `main.tf`. See https://github.com/dfds/infrastructure-modules/blob/master/compute/k8s-services/versions.tf for reference.

wcarlsen:
@dingobar do you still have interest in this or should I close the PR?

PR #198, opened 2021-04-13
Allow selecting db instance class instead of defaulting to t2 micro

dingobar:
Because:
- Sometimes a larger instance is needed
- t2 is already previous generation

### DIFF
--- a/database/postgres/vars.tf
+++ b/database/postgres/vars.tf
@@ -38,3 +38,8 @@ variable "engine_version" {
   description = "RDS engine version"
   default     = null
 }
+variable "db_instance_class" {
+  type        = string
+  description = "RDS database instance class"
+  default     = "db.t2.micro"
+}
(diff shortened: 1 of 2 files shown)


database/postgres/vars.tf
```diff
@@ -37,4 +37,9 @@ variable "engine_version" {
   type        = string
   description = "RDS engine version"
   default     = null
-}
\ No newline at end of file
+}
+variable "db_instance_class" {
+  type        = string
+  description = "RDS database instance class"
+  default     = null
```
wcarlsen, line 44: This breaks the module, compared to previous usage. If you checkout the _sub/database/postgres submodule you can see that the default is "db.t2.micro", so your default should also reflect this if you don't want to break stuff.

wcarlsen (changes requested):
Overall good ida for a change. Please consider backwards compatibility as noted in the comment.

wcarlsen:
https://github.com/dfds/infrastructure-modules/issues/199

database/postgres/vars.tf
dingobar: Fixed, default now the same as in the submodule.

wcarlsen (approved):

PR #200, opened 2021-04-13
Add parameter family variable

dingobar:
# UPDATE

After a quick chat, we decided to
1. Use the officially supported terraform rds module if possible
2. Change the approach taken to this feature to string concatenation in stead of anoter variable

--------------

Hardcoded parameter group family makes it impossible to use a db engine version >10. Therefore, I have added another parameter. The change is backwards compatible because the default is still postgres10. I've observed that engine_version is not supplied, terraform will default to the latest version, unless the parameter group is not compatible, in which case it will go for the newest version compatible.

For my use case, I will be ommitting the engine_version variable, but supply the parameter family variable. This will keep my database at the latest compatible version.

I have only manually tested this with this configuration:
```HCL
module "database" {
  source                    = "git::https://github.com/dingobar/infrastructure-modules.git//database/postgres?ref=780afa09a2d9d2674b878703fd708c2f33536e69"
  db_name                   = "mlflow${var.environment[terraform.workspace]}"
  aws_region                = data.aws_region.current.name
  application               = "dfds-mlflow"
  db_master_username        = "superadmin"
  db_master_password        = random_password.rds_master_password.result
  skip_final_snapshot       = var.skip_snapshot[terraform.workspace]
  db_instance_class         = "db.t3.micro"
  db_parameter_group_family = "postgres12"
}
```
as well as with no `db_parameter_group_family`.

Changing `db_parameter_group_family` causes the whole parameter group to be destroyed and re-created, which probably is not an issue.
```sh
  # module.database.module.postgres.aws_db_parameter_group.dbparams must be replaced
-/+ resource "aws_db_parameter_group" "dbparams" {
      [...]
      ~ description = "Force SSL encryption for postgres10" -> "Force SSL encryption for postgres13" # forces replacement
      ~ family      = "postgres10" -> "postgres13" # forces replacement
      [...]
      ~ name        = "dfds-mlflow-postgres10-force-ssl-prod" -> "dfds-mlflow-postgres13-force-ssl-prod" # forces replacement
    }
```

### DIFF
--- a/_sub/database/postgres/main.tf
+++ b/_sub/database/postgres/main.tf
@@ -1,3 +1,6 @@
+locals {
+  engine_family = "${var.engine}${var.engine_version}"
+}
 resource "aws_security_group" "pgsg" {
   name_prefix = "${var.application}-postgres10-sg-${var.environment}"
   description = "Allow all inbound traffic on port ${var.db_port}"
@@ -21,9 +24,9 @@ resource "aws_security_group" "pgsg" {
 
 #Enable SSL on the database by default
 resource "aws_db_parameter_group" "dbparams" {
-  name        = "${var.application}-postgres10-force-ssl-${var.environment}"
-  description = "Force SSL encryption for postgres10"
-  family      = "postgres10"
+  name        = "${var.application}-${var.engine_family}-force-ssl-${var.environment}"
+  description = "Force SSL encryption for ${var.engine_family}"
+  family      = var.engine_family
 
   parameter {
     name  = "rds.force_ssl"
--- a/_sub/database/postgres/main.tf
+++ b/_sub/database/postgres/main.tf
@@ -1,5 +1,5 @@
 locals {
-  engine_family = "${var.engine}${var.engine_version}"
+  engine_family = "${aws_db_instance.postgres.engine}${var.engine_version}"
 }
 resource "aws_security_group" "pgsg" {
   name_prefix = "${var.application}-postgres10-sg-${var.environment}"
--- a/_sub/database/postgres/main.tf
+++ b/_sub/database/postgres/main.tf
@@ -1,8 +1,8 @@
 locals {
-  engine_family = "${aws_db_instance.postgres.engine}${var.engine_version}"
+  engine_family = var.engine_version == null ? "postgres13" : "postgres${var.engine_version}"
 }
 resource "aws_security_group" "pgsg" {
-  name_prefix = "${var.application}-postgres10-sg-${var.environment}"
+  name_prefix = "${var.application}-postgres-sg-${var.environment}"
   description = "Allow all inbound traffic on port ${var.db_port}"
 
   ingress {
@@ -41,7 +41,7 @@ resource "aws_db_parameter_group" "dbparams" {
 #Restore the postgres database with the pre-configured settings
 resource "aws_db_instance" "postgres" {
   engine                  = "postgres"
-  engine_version          = var.engine_version != null ? var.engine_version : null
+  engine_version          = var.engine_version
   publicly_accessible     = "true"
   backup_retention_period = 10
   apply_immediately       = true
--- a/_sub/database/postgres/vars.tf
+++ b/_sub/database/postgres/vars.tf
@@ -46,10 +46,6 @@ variable "skip_final_snapshot" {
 
 variable "engine_version" {
   type        = string
-  description = "RDS engine version"
-  default     = 10
-  validation {
-    condition     = can(regex("^[0-9]+$", var.engine_version))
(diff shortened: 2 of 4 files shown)


_sub/database/postgres/main.tf
```diff
@@ -21,9 +21,9 @@ resource "aws_security_group" "pgsg" {
 
 #Enable SSL on the database by default
 resource "aws_db_parameter_group" "dbparams" {
-  name        = "${var.application}-postgres10-force-ssl-${var.environment}"
-  description = "Force SSL encryption for postgres10"
-  family      = "postgres10"
+  name        = "${var.application}-${var.db_parameter_group_family}-force-ssl-${var.environment}"
+  description = "Force SSL encryption for ${var.db_parameter_group_family}"
+  family      = var.db_parameter_group_family
```
wcarlsen, line 26: No need to introduce a new variable that is already a combination of var.engine and var.engine_version. Instead create a locals and combine the two variables into a engine_familiy variable.

wcarlsen: Also you forgot about the naming of the security group above in line 2.

_sub/database/postgres/vars.tf
```diff
@@ -39,6 +39,11 @@ variable "db_allocated_storage" {
   default     = 20
 }
 
+variable "db_parameter_group_family" {
+  description = "Postgresql parameter group family compatible with chosen engine version"
+  default     = "postgres10"
```
wcarlsen, line 44: Still this variable should not be needed, since it is a combination of engine and engine_version

database/postgres/README.md
```diff
@@ -1,12 +1,16 @@
...
-* db_master_password = "testtesttest" (Can be used for changing password)
-* skip_final_snapshot = false (Optional and should NOT be set to true for production)
\ No newline at end of file
+
+- db_name = "papp" (Destructive if changed)
+- aws_region = "eu-central-1" (Destructive if changed)
+- application = "application" (Destrutive if changed)
+- db_master_username = "test" (Destrutive if changed)
+- db_master_password = "testtesttest" (Can be used for changing password)
+- db_parameter_group_family = "postgres13" (Optional, but default is postgres10)
```
wcarlsen, line 14: Remove this.

database/postgres/main.tf
```diff
@@ -20,16 +20,17 @@ provider "aws" {
...
+  application               = var.application
+  environment               = var.environment
+  db_name                   = var.db_name
+  db_master_username        = var.db_master_username
+  db_master_password        = var.db_master_password
+  db_port                   = var.db_port
+  skip_final_snapshot       = var.skip_final_snapshot
+  engine_version            = var.engine_version
+  db_instance_class         = var.db_instance_class
+  db_parameter_group_family = var.db_parameter_group_family
```
wcarlsen, line 33: remove.

database/postgres/vars.tf
```diff
@@ -43,3 +43,8 @@ variable "db_instance_class" {
   description = "RDS database instance class"
   default     = "db.t2.micro"
 }
+
+variable "db_parameter_group_family" {
```
wcarlsen, line 47: This is again not needed.

wcarlsen (changes requested):
Overall a good PR. Minor changes is needed. Let's run a test together maybe, because I think this has not been testet?

Check here to about locals in Terraform https://www.terraform.io/docs/language/values/locals.html

_sub/database/postgres/main.tf
dingobar: Engine version can also include a minor version, so by doing it this way we restrict ourselves from using those, ie. engine_version 13 will be ok, but not 13.1 if we ever had to run that specific version (although the default is to allow minor version upgrades, so 13 automatically means latest minor 13 version). Are we ok with that? (I am)

Another thing, how certain are we that the family name will always be `${engine}${engine_version}`? This way of doing things will break if AWS brings out family postgres13.5 or something like that, which they are doing on mysql. Looks like the postgresql family names have been like this since v11, so maybe we can assume it wil stay this way?

If we are ok with those two limitations, I will go ahead and refactor the PR with the approach you outlined.

dingobar: About the security group, it does not have anything to do with the engine version in the first place, so I would actually prefer either just renaming it from `postgres10` to `postgres`, or just leaving it as it is.

dingobar: Resolved after call

wcarlsen (approved):
LGTM! But would like an external reviewer.

wcarlsen:
Would like an extra set of eyes on it.

rifisdfds (approved):

PR #202, opened 2021-04-20
Velero s3

wcarlsen:


dfds-pwes (commented):
A couple of things.

1. We seem to be missing the destroy-cluster step for k8s-qa19 in the destroy_eks_1_19 job.
2. Missing variable descriptions for some variables but not sure if these are just ommmitted when the variable name is explicit enough to warrant them not being required.

Also, i'm curious about the change made in the postgres module.  I presume that is totally unrelated to Velero?

dfds-pwes (approved):
LGTM

PR #203, opened 2021-04-22
Add cloudwatch iam role for grafana

rifisdfds:
This PR creates an IAM role with the required permissions to retrieve Cloudwatch metrics. It annotates the Grafana pods and monitoring namespace which is a prerequisite to allow us to remove the IAM key/secret currently set in the Cloudwatch datasource in our Grafana deployment and replace it with the SDK option which will automatically assume the created role via the pod annotation applied

### DIFF
--- a/_sub/compute/helm-kube-prometheus-stack/vars.tf
+++ b/_sub/compute/helm-kube-prometheus-stack/vars.tf
@@ -44,6 +44,11 @@ variable "grafana_notifier_name" {
   description = "Grafana notifier name"
 }
 
+variable "grafana_iam_role_name" {
+  type        = string
+  description = "Name to be given to the Grafana IAM role"
+}
+
 variable "slack_webhook" {
   type        = string
   description = "Alert Slack webhook"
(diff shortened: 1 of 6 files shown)


wcarlsen (dismissed):
LGTM!

_sub/compute/helm-kube-prometheus-stack/vars.tf
```diff
@@ -1,3 +1,11 @@
+variable "aws_workload_account_id" {
```
avnes, line 1: No type definition for this one?

avnes (changes requested):

_sub/compute/helm-kube-prometheus-stack/vars.tf
rifisdfds: There's no type defined for it at the level it's passed in from (compute/k8s-services), and I kept it consistent with the _sub\compute\k8s-kiam submodule where it's also used in the same way.

That's not to say it shouldn't have a type set, or a description for that matter. I'll change it

avnes (approved):

avnes (approved):

PR #204, opened 2021-04-23
Retire deprecated whitelist option for kiam v4

avnes:
NOT TO BE MERGED UNTIL next week.

### DIFF
--- a/_sub/compute/k8s-kiam/vars.tf
+++ b/_sub/compute/k8s-kiam/vars.tf
@@ -91,7 +91,7 @@ variable "chart_version" {
   default     = null
 
   validation {
-    condition     = tonumber(substr(var.chart_version, 0, 1)) >= 6
+    condition     = tonumber(split(".", var.chart_version)[0]) >= 6
     error_message = "The KIAM Helm chart_version must be greater than 6.0.0."
   }
 }
--- a/_sub/compute/k8s-kiam/vars.tf
+++ b/_sub/compute/k8s-kiam/vars.tf
@@ -91,7 +91,7 @@ variable "chart_version" {
   default     = null
 
   validation {
-    condition     = tonumber(split(".", var.chart_version)[0]) >= 6
+    condition     = tonumber(element(split(".", var.chart_version), 0)) >= 6
     error_message = "The KIAM Helm chart_version must be greater than 6.0.0."
   }
 }
--- a/_sub/compute/k8s-kiam/vars.tf
+++ b/_sub/compute/k8s-kiam/vars.tf
@@ -89,6 +89,11 @@ variable "chart_version" {
   type        = string
   description = "KIAM helm chart version"
   default     = null
+
+  validation {
+    condition     = tonumber(substr(var.chart_version, 0, 1)) >= 6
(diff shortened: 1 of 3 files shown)


abstrask (changes requested):
With this change, we only support `chart_version` 6.0.0 and above.

Maybe we should configure validation for this variable in `vars.tf``, to avoid the problem noted in the EKS manifest for our cluster:

> 5.10.0 includes https://github.com/uswitch/kiam/pull/427, but image does not - breaks WhitelistRouteRegexp/AllowRouteRegexp

wcarlsen (approved):
Nice with the validate part.

avnes:
> With this change, we only support `chart_version` 6.0.0 and above.
> 
> Maybe we should configure validation for this variable in `vars.tf``, to avoid the problem noted in the EKS manifest for our cluster:
> 
> > 5.10.0 includes [uswitch/kiam#427](https://github.com/uswitch/kiam/pull/427), but image does not - breaks WhitelistRouteRegexp/AllowRouteRegexp

That is a good suggestion, so I have implemented that now. I also tested if it worked by revirterint to the old version (5.9.0), and the validation kicked in:

Screenshot 2021-04-26 at 09 33 28

_sub/compute/k8s-kiam/vars.tf
```diff
@@ -89,6 +89,11 @@ variable "chart_version" {
   type        = string
   description = "KIAM helm chart version"
   default     = null
+
+  validation {
+    condition     = tonumber(substr(var.chart_version, 0, 1)) >= 6
```
wcarlsen, line 94: This wont work with versions greater than single digit

dfds-pwes: Presume a split function call might be a good idea?  Should turn the chart version into a list containing major, minor and patch versions.  Then the comparison can just be done on the major portion.

abstrask, line 94: `condition = element(split(".", var.chart_version), 0) >= 6` should take future proof it

abstrask (changes requested):

_sub/compute/k8s-kiam/vars.tf
dfds-pwes: Question though.  If you use a split command on a string then do the returned elements come back as strings or integers?  Is there not a need to cast data types here?

abstrask: It will be implicitly converted, I think. I verified with this code:

```hcl
locals {
  version = "6.1.2"
  first = element(split(".", local.version), 0)
}

output "ge-6" {
  value = local.first >= 6
}

output "ge-7" {
  value = local.first >= 7
}
```

avnes: Implemented as:

`tonumber(element(split(".", var.chart_version), 0)) >= 6`

SEQUOIIA:
@wcarlsen

wcarlsen (approved):
LGTM!

wcarlsen (approved):
LGTM!

avnes:
@raras-dfds - could you also please take a second look?

abstrask (approved):

PR #205, opened 2021-04-26
moved grafana iam role to own module

rifisdfds:
This PR creates the IAM role for grafana outside of kube-prometheus-stack to remove the circular dependency between kube-prometheus-stack and kiam that occurs


dfds-pwes (approved):
Looks good to me.

PR #209, opened 2021-04-30
Add crossplane k8s service for prom stack

rifisdfds:
This PR adds 2 Kubernetes services for crossplane labeled so that the metrics provided by the pods can be scraped by the kube-prometheus-stack deployment


dfds-pwes (approved):
LGTM

PR #211, opened 2021-05-03
Use ARN instead of name for ebs-csi-driver role name

avnes:


wcarlsen (approved):
Making it work with KIAM strict mode. NICE!

PR #213, opened 2021-05-03
Fix qa

abstrask:


src/qa-test-eks.sh
```diff
@@ -86,26 +118,22 @@ if [ "$ACTION" = "test" ]; then
...
     REGION=$2
     CLUSTERNAME=$3
-    aws --region $REGION eks update-cluster-config --name $CLUSTERNAME --logging '{"clusterLogging":[{"types":["api","audit","authenticator","controllerManager","scheduler"],"enabled":false}]}' || true
-    sleep 60
-fi
+    WORKDIR="${BASEPATH}/${REGION}/k8s-${CLUSTERNAME}"
 
+    # Disable cluster logging
+    echo Disabling cluster logging
+    aws --region "$REGION" eks update-cluster-config --name "$CLUSTERNAME" --logging '{"clusterLogging":[{"types":["api","audit","authenticator","controllerManager","scheduler"],"enabled":false}]}' || true
```
avnes, line 129: What is the reason why you use "OR true". Do you expect this command to fail under certain circumstances?

avnes (changes requested):

src/qa-test-eks.sh
abstrask: Good questions. Probably out of an abundance of caution, more than this command being particular error prone.

Previously, we decided we want everything *but* the log groups to get cleaned up, if this step happened to fail. But we could easily consider failing the pipeline if this step fails.

Thoughts?

abstrask: FWIW, the `|| true` was there before was well. I've just merged this step into the destroy job, instead of a separate.

avnes: It's probably best to keep is as-is then, since it would be scope screep to revisit the reason for this OR clause now.

avnes (approved):

PR #216, opened 2021-05-06
Make Azure AppReg sub module entirely Terraform native

abstrask:
As a consequence, we can remove Azure CLI from our pipline image, saving ~500 MB.


abstrask:
Hmm, I seem to have left out the `required_resource_access` config block. 

Need to consider if that needs to be added back in before merging.

abstrask:
I mean this block specifically:

```hcl
required_resource_access {
  resource_app_id = "00000002-0000-0000-c000-000000000000"

  resource_access {
    id   = local.SIGN_IN_AND_READ_USER_PROFILE
    type = "Scope"
  }
} 
```

But `00000002-0000-0000-c000-000000000000` is the client Id of Azure AD Graph API, and that whole block was only used by the `aad_access` "leg" of the module (which does not seem to be used anywhere).

The `no_aad_access`, on the other hand, did not reference this at all, so I think we're good to go as is!

rifisdfds:
Running the branch stops sandbox from being able to deploy services. Are there any special requirements/changes needed?

Previously I would run the `az login --service-principal --username "$ARM_CLIENT_ID" --password "$ARM_CLIENT_SECRET" --tenant "$ARM_TENANT_ID"` step manually before applying services. Now that does not seem to work anymore

abstrask:
`az login` no longer required - it should work fine without :)

rifisdfds:
To clarify, I tried `az login` after initially trying without

rifisdfds:
It seems to be an issue when `traefik_alb_auth_deploy = false` is set, in particular

abstrask:
@avnes @rifisdfds Think it's fixed now. Can you verify?

rifisdfds:
Though that may be because my previous login via `az login` was cached. `az logout` has cleared it and I now get:

```
Error: Building AzureAD Client

obtain tenant() from Azure CLI: Error parsing json result from the Azure CLI:
Error waiting for the Azure CLI: exit status 1: ERROR: Please run 'az login'
to setup account.
```

abstrask:
Maybe delete `.terragrunt-cache` and try again?

rifisdfds:
@raras-dfds I was not exporting them 2nd time round actually.

If the alb auth is set to false, should we have a conditional for the use of those values where they don't need to be provided? I'm not sure of any knock-on effects of that so this is a question rather than a suggestion

abstrask:
I wouldn't think the provider would be initialised, and therefore that the ARM variables necessary, if the module has a count of zero. Will test...

dfds-pwes:
I had some fun and games with this because my sandbox was quite a long way behind.  I was missing both the bump to 1.19 (not really the essential) but more importantly some of the roles referenced in the services file now need the full ARN and not just the role name.

I've got my sandbox up to date now so I can try it without any of the az login stuff to see if it runs through if I target the azure-appreg-v2 branch now.   Just for the sake of comparison.

dfds-pwes:
Just tested with a newly spun up docker image to make sure that the environment variables weren't present.  I also cleared the terragrunt cache but I still get the same error as Richard.

This error:

`Error: Building AzureAD Client

Authenticating using the Azure CLI is only supported as a User (not a Service
Principal).

To authenticate to Azure using a Service Principal, you can use the separate
'Authenticate using a Service Principal'
auth method - instructions for which can be found here:

Alternatively you can authenticate using the Azure CLI by using a User
Account.`

wcarlsen (dismissed):
LGTM!

dfds-pwes:
We did some investigation into this during today and below are our notes which will be required for the migration process.  Note that the migration will result in a short outage on the Auth ALB.

1. Going forward you will need the environment variables ARM_TENANT_ID, ARM_CLIENT_ID and ARM_CLIENT_SECRET to be presented at the point of deployment.  For EKS-Pipeline the variables are already provided as part of the pipeline, but if you are running interactively in a Docker container then you will need to tear down your currently defined container and redeploy with these variables provided.  It means a small amount of work but removes the need for you to do the 'az login' each time you spin up a new instance of the container.

2. If you receive an error regarding an AD Application which already exists then you can use the following commands to remove the existing entry.

```
# query the app list for the specific identifier-uri required
az ad app list --identifier-uri https://xxxx.yyyy.dfds.cloud

# delete the existing application (take the appId from the output from the previous command)
az ad app delete --id [AppId]
```

@abstrask I've commented here but will leave it for you to sort out the conflicts and do the squash and merge as I guess it will need some coordination and probably not something to do on a Friday so late in the day.

abstrask:
> 
> 
> I'm going to bump the AzureAD provider version from 1.2.2 to 1.5.1 so we either need to rebase this from master or also make the same change here.

Can you bump it on this branch then? Thanks! :)

abstrask:
Migration:

- Might fail to re-create Azure AppReg due to sequencing. Solution: Run again
- Authentication will be out of service for a few minutes, between old appreg deleted, and new one succesfully created
- Users will be prompted to trust new Azure application
- Scheduled maintenance outside of biz hours

Otherwise good to go!

dfds-pwes (approved):
Looks good to me.

PR #220, opened 2021-05-17
Add velero role cleanup to qa script

rifisdfds:
- Provide missing region to the cleanup action
- Remove unnecessary parameters from extra_cleanup function
- Added "Velero" prefix to list of IAM roles cleaned up

### DIFF
--- a/azure-pipelines.yaml
+++ b/azure-pipelines.yaml
@@ -50,7 +50,7 @@ stages:
         steps:
           - bash: ./src/qa-test-eks.sh init
             displayName: "Init"
-          - bash: ./src/qa-test-eks.sh cleanup qa19
+          - bash: ./src/qa-test-eks.sh cleanup eu-west-1 qa19
             displayName: "Pre-apply cleanup"
             condition: eq(variables['PRE_CLEANUP'], 'true')
           - bash: ./src/qa-test-eks.sh plan-cluster eu-west-1 qa19
--- a/src/qa-test-eks.sh
+++ b/src/qa-test-eks.sh
@@ -6,8 +6,8 @@ ACTION=$1
 # $AWS_DEFAULT_REGION
 
 extra_cleanup () {
-    REGION=$1
-    export CLUSTERNAME=$2
+    REGION=$2
+    CLUSTERNAME=$3
 
     # Remove specific resources that sometimes get left behind (always return true, as resource may have been successfully been cleaned up)
     aws --region "$REGION" iam list-roles --output json | jq -r --arg ROLEPREFIX "eks-${CLUSTERNAME}-" '.Roles[] | select( .RoleName | contains($ROLEPREFIX, "Velero") ) | .RoleName' | xargs -r -L1 aws --region "$REGION" iam delete-role --role-name || true
--- a/src/qa-test-eks.sh
+++ b/src/qa-test-eks.sh
@@ -5,9 +5,7 @@ BASEPATH=./test/integration
 ACTION=$1
 # $AWS_DEFAULT_REGION
 
-extra_cleanup () {
-    REGION=$2
-    CLUSTERNAME=$3
+function extra_cleanup {
 
     # Remove specific resources that sometimes get left behind (always return true, as resource may have been successfully been cleaned up)
     aws --region "$REGION" iam list-roles --output json | jq -r --arg ROLEPREFIX "eks-${CLUSTERNAME}-" '.Roles[] | select( .RoleName | contains($ROLEPREFIX, "Velero") ) | .RoleName' | xargs -r -L1 aws --region "$REGION" iam delete-role --role-name || true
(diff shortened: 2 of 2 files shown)


src/qa-test-eks.sh
```diff
@@ -5,12 +5,10 @@ BASEPATH=./test/integration
 ACTION=$1
 # $AWS_DEFAULT_REGION
 
-extra_cleanup () {
-    REGION=$1
-    export CLUSTERNAME=$2
+function extra_cleanup {
```
abstrask, line 8: The JQ program needs the `CLUSTERNAME` variable exported (not just local variables). Also, region is remove from this function - are we sure it's inherited from where it's called?

src/qa-test-eks.sh
```diff
@@ -36,7 +34,7 @@ if [ "$ACTION" = "cleanup" ]; then
     CLUSTERNAME=$3
 
     # Remove specific resources that sometimes get left behind (always return true, as resource may have been successfully been cleaned up)
-    extra_cleanup "$REGION" "$CLUSTERNAME"
+    extra_cleanup
```
abstrask, line 37: Depending on outcome of comment above, the arguments may need to be reinstated.

src/qa-test-eks.sh
```diff
@@ -133,7 +131,7 @@ if [ "$ACTION" = "destroy-cluster" ]; then
     terragrunt destroy-all --terragrunt-working-dir "$WORKDIR" --terragrunt-source-update --terragrunt-non-interactive -input=false -auto-approve || RETURN=1
     
     # Remove specific resources that sometimes get left behind (always return true, as resource may have been successfully been cleaned up)
-    extra_cleanup "$REGION" "$CLUSTERNAME"
+    extra_cleanup
```
abstrask, line 134: Depending on outcome of comment above, the arguments may need to be reinstated.

src/qa-test-eks.sh
rifisdfds: I could be missing something but why is the export required? We are not using local variables in the bash script, just global? The region should be set correctly each the function is called so it's not needed to set that or the clustername again at this point

azure-pipelines.yaml
```diff
@@ -50,7 +50,7 @@ stages:
         steps:
           - bash: ./src/qa-test-eks.sh init
             displayName: "Init"
-          - bash: ./src/qa-test-eks.sh cleanup qa19
+          - bash: ./src/qa-test-eks.sh cleanup eu-west-1 qa19
             displayName: "Pre-apply cleanup"
```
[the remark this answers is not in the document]
rifisdfds: I think it's useful if there's a resource left over from a previous run. You can re-run the pipeline with the variable set and it will clean it up (provided it is found by the query of course). The alternative is to do a manual clean up?

src/qa-test-eks.sh
abstrask: JQ being weird. It was first fixed in this commit: https://github.com/dfds/infrastructure-modules/commit/9c55728a00cc78d44d66665bfd652262469c56cc.

If it's only a local variable, and not exported, the argument is not passed (via `--arg ROLEPREFIX "eks-${CLUSTERNAME}-"`) properly to the JQ program, meaning it won't match the roles.

I haven't tested if REGION is inherited in the extra_cleanup function.

azure-pipelines.yaml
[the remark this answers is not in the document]
abstrask: Sounds reasonable.

src/qa-test-eks.sh
rifisdfds: I'm not seeing that happen. To confirm I've just ran a test by creating an eks-qa19-rifistest role and running pre-cleanup via pipeline. It picks up that the role exists and tries to delete it so that indicates to me that it's working without needing to export.

Unfortunately that test appears to have shown a different problem:

```
An error occurred (DeleteConflict) when calling the DeleteRole operation: Cannot delete entity, must remove roles from instance profile first.
```

Which may require some extra changes to the cleanup action. I can't say why me creating that role in the web console has created an instanceprofile, mind, but one also gets created for eks-qa19-node so I'm guessing it's an AWS Service EC2 trusted entity thing. One for another pull request if it happens for real, perhaps, rather than scope creep with this one

abstrask: The JQ query consistenly failed for us, until we specifically exported the variable used with `--arg`, as @wcarlsen is my witness 👀. Not sure what have changed.

rifisdfds: The Prime-pipeline docker image was recently updated which would have picked up any updates to `jq` if there have been any in the base image

All I know for sure is that it's consistently working now (I've ran yet another test with no InstanceProfile or Policy attached to the role and it has been deleted by pre-cleanup) and variables are set as expected.

I'd like to merge this in based on the above discussion and test if we are happy this can be marked as resolved? I'll raise a new issue for the modification to include `remove-role-from-instance-profile` and `detach-role-policy` if any are present on the roles found

abstrask (approved):

PR #221, opened 2021-05-19
Ingress to Traefik. Secured by Basic Auth

avnes:
**Requirements in services/terragrunt.hcl:**

```
traefik_alb_auth_deploy     = true
traefik_dashboard_deploy    = true
```

For production also another DNS alias is required:
`traefik_alb_auth_core_alias = ["build.dfds.cloud", "traefik.dfds.cloud"]`

**Logic:**

When traefik_dashboard_deploy is true, the following will happen:

- A random password will be generated. 32 characters long. Contains alphanumeric characters and symbols.
- This password will be stored as a secure string in the AWS Parameter Store.
- The password will be hashed. 
- Together with a username the hashed value of the password will be stored in a Kubernetes secret.
- This means that the clear text representation of the password can never be retrieved from the secret; only from Parameter Store.
- This Kubernetes secret is used to add Basic Authentication to a new Ingress into the Traefik dashboard.
- The new ingress to access the Traefik dashboard will also be protected by the authenticated ALB.
- The hostname used by the ingress depends on this being a production environment or not. The hostname logic is explained below.

**Ingress hostname logic:**

If the traefik_alb_auth_core_alias has an list item with the value _traefik.dfds.com_ (like it will have in production once the eks-pipeline is updated), the ingress hostname will be _traefik.dfds.com_.

Else (for other environments), the ingress hostname will be internal.cluster-name.capability-name.dfds.cloud

### DIFF
--- a/_sub/compute/k8s-traefik/vars.tf
+++ b/_sub/compute/k8s-traefik/vars.tf
@@ -84,8 +84,7 @@ variable "dashboard_ingress_backend_path" {
   default     = "/"
 }
 
-variable "dashboard_ingress_rule_host" {
+variable "dns_zone_name" {
   type        = string
-  description = "The hostname to access the Traefik dashboard on."
-  default     = null
+  description = "The dns_zone_name. For sandbox this is including the identity name."
 }
--- a/_sub/compute/k8s-traefik/vars.tf
+++ b/_sub/compute/k8s-traefik/vars.tf
@@ -46,6 +46,12 @@ variable "request_memory" {
   default = "128Mi"
 }
 
+variable "dashboard_deploy" {
+  type        = bool
+  description = "Deploy ingress for secure access to Traefik dashboard."
+  default     = false
+}
+
 variable "dashboard_username" {
   type        = string
   description = "Username used for basic authentication."
--- a/_sub/compute/k8s-traefik/vars.tf
+++ b/_sub/compute/k8s-traefik/vars.tf
@@ -49,7 +49,7 @@ variable "request_memory" {
 variable "dashboard_deploy" {
   type        = bool
   description = "Deploy ingress for secure access to Traefik dashboard."
-  default     = false
+  default     = true
 }
 
 variable "dashboard_username" {
--- a/_sub/compute/k8s-traefik/vars.tf
+++ b/_sub/compute/k8s-traefik/vars.tf
@@ -45,3 +45,41 @@ variable "request_memory" {
   description = "(optional) Describes the minimum amount of memory required"
   default = "128Mi"
 }
+
+variable "dashboard_username" {
+  type        = string
+  description = "Username used for basic authentication."
+  default     = "cloudengineer"
+}
+
+variable "dashboard_secret_name" {
+  type        = string
+  description = "Name of the k8s secret to store the credentials for basic authentication."
+  default     = "traefik-basic-auth"
+}
+
+variable "dashboard_ingress_name" {
+  type        = string
+  description = "Name of the ingress, must be unique."
+  default     = "traefik-dashboard"
+}
+
+variable "dashboard_ingress_labels" {
+  type        = map(string)
+  description = "Map of string keys and values that can be used to organize and categorize the ingress."
+  default     = {
+    "name" = "traefik-dashboard"
+  }
+}
+
+variable "dashboard_ingress_backend_path" {
+  type        = string
+  description = "The path for the service entry point."
+  default     = "/"
+}
+
+variable "dashboard_ingress_rule_host" {
+  type        = string
+  description = "The hostname to access the Traefik dashboard on."
+  default     = null
+}
--- a/_sub/compute/k8s-traefik/vars.tf
+++ b/_sub/compute/k8s-traefik/vars.tf
@@ -84,7 +84,7 @@ variable "dashboard_ingress_backend_path" {
   default     = "/"
 }
 
-variable "dns_zone_name" {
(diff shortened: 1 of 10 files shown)


compute/k8s-services/vars.tf
```diff
@@ -129,6 +134,12 @@ variable "traefik_health_check_path" {
   default = "/dashboard/"
 }
 
+variable "traefik_dashboard_deploy" {
+  type        = bool
+  description = "Deploy ingress for secure access to Traefik dashboard."
+  default     = false
```
abstrask, line 140: Should we set default to "true"?

compute/k8s-services/vars.tf
```diff
@@ -67,6 +67,11 @@ variable "cloudwatchlogs_iam_role_deploy" {
 # Traefik
 # --------------------------------------------------
 
+variable "traefik_svc_name" {
+  type    = string
+  default = "traefik"
```
abstrask, line 72: Unless we are explicit about service names, this will clash with multiple instances. Consider setting service name to just deploy_name, and not have a separate variable?

Or default to null, and configure the `kubernetes_service` resource to use supplied service name, or fallback to deploy_name if null?

compute/k8s-services/dependencies.tf
```diff
@@ -197,3 +197,20 @@ data "aws_iam_policy_document" "cloudwatch_metrics_trust" {
     actions = ["sts:AssumeRole"]
   }
 }
+
+# --------------------------------------------------
+# Traefik dashboard secure access
+# --------------------------------------------------
+
+locals {
```
abstrask, line 205: Can we add some comments on what this logic does? At least it wasn't obvious to my afternoon brain :)

_sub/compute/k8s-traefik/vars.tf
```diff
@@ -44,4 +44,47 @@ variable "request_memory" {
...
   description = "(optional) Describes the minimum amount of memory required"
   default = "128Mi"
-}
\ No newline at end of file
+}
+
+variable "dashboard_deploy" {
+  type        = bool
+  description = "Deploy ingress for secure access to Traefik dashboard."
+  default     = false
```
abstrask, line 52: Would it make sense to default to true?

_sub/compute/k8s-traefik/vars.tf
```diff
@@ -44,4 +44,47 @@ variable "request_memory" {
...
+variable "dashboard_username" {
+  type        = string
+  description = "Username used for basic authentication."
+  default     = "cloudengineer"
+}
+
+variable "dashboard_secret_name" {
+  type        = string
+  description = "Name of the k8s secret to store the credentials for basic authentication."
+  default     = "traefik-basic-auth"
```
abstrask, line 64: This risk clashing with additional Traefik instances. Should this contain the deploy_name, e.g. `${var.deploy_name}-basic-auth`?

_sub/compute/k8s-traefik/vars.tf
```diff
@@ -44,4 +44,47 @@ variable "request_memory" {
...
+}
+
+variable "dashboard_ingress_name" {
+  type        = string
+  description = "Name of the ingress, must be unique."
+  default     = "traefik-dashboard"
```
abstrask, line 70: This risk clashing with additional Traefik instances. Should this contain the deploy_name, e.g. ${var.deploy_name}-dashboard?

abstrask (commented):
Good job. Added a few comments and questions for clarification.

compute/k8s-services/vars.tf
avnes: I am not sure I understand this suggestion. The traefik_deploy module was already hardcoding "traefik" as deploy_name. So this was partially to avoid hardcoding, but more importantly to also use the variable traefik_svc_name in conditional statements to determine locals defined in dependencies.tf

compute/k8s-services/dependencies.tf
avnes: Yes, that is understandable :)

I added the explaination in the PR details:

Ingress hostname logic:

If the traefik_alb_auth_core_alias has an list item with the value traefik.dfds.com (like it will have in production once the eks-pipeline is updated), the ingress hostname will be traefik.dfds.com.

Else (for other environments), the ingress hostname will be internal.cluster-name.capability-name.dfds.cloud

compute/k8s-services/vars.tf
avnes: Fixed.

_sub/compute/k8s-traefik/vars.tf
avnes: Fixed.

_sub/compute/k8s-traefik/vars.tf
avnes: Fixed.

_sub/compute/k8s-traefik/vars.tf
avnes: Fixed.

compute/k8s-services/vars.tf
avnes: Also keep in mind that there is no such variable called deploy_name in k8s-services, so the traefik_svc_name variable is passed down to the k8s-traefik sub module's deploy_name variable.

avnes: Added documentation for the "weird" logic on how to calculate the traefik dashboard hostname. And also remove the traefik_svc_name variable.

abstrask (approved):
LGTM!

PR #223, opened 2021-05-27
Crossplane chart now installs providers via manifests

rifisdfds:
This PR changes the installation of the crossplane providers to a manifest instead of via the helm chart. This will prevent a problem upgrading from 1.0.0 to 1.2.1 caused by the newer chart using a new naming convention for the providers it installs, and will also prevent future upgrade problems which deploy upgraded provider versions as brand new providers, removing the previous ones

### DIFF
--- a/_sub/compute/helm-crossplane/versions.tf
+++ b/_sub/compute/helm-crossplane/versions.tf
@@ -1,3 +1,10 @@
 terraform {
   required_version = ">= 0.12"
+
+  required_providers {
+
+    kubectl = {
+      source  = "gavinbunney/kubectl"
+    }
+  }
 }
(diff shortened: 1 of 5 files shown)


_sub/compute/helm-crossplane/versions.tf
```diff
@@ -1,3 +1,10 @@
 terraform {
   required_version = ">= 0.12"
+
+  required_providers {
+
+    kubectl = {
+      source  = "gavinbunney/kubectl"
```
ZaradarDFDS, line 7: This could be a potential RAT carrier. But as such I recon you guys trust Gavin :)

ZaradarDFDS (approved):

_sub/compute/helm-crossplane/versions.tf
rifisdfds: Yeah I thought the same but we're using it in here already and all seems legitimate

abstrask: @ZaradarDFDS More like RAbbiT!

PR #225, opened 2021-05-31
Eks qa detach policies

abstrask:
- Remove pipefail option
- Disable CLI pager
- Detach node role policies before delete
- Print xargs command before executing
- Separate cleanup functions for cluster and shared resources
- Fix unbound variable 

Results:
https://dev.azure.com/dfds/CloudEngineering/_build/results?buildId=367705&view=results
https://dev.azure.com/dfds/CloudEngineering/_build/results?buildId=367707&view=results

### DIFF
--- a/src/qa-test-eks.sh
+++ b/src/qa-test-eks.sh
@@ -10,13 +10,13 @@ function extra_cleanup {
     # Remove specific resources that sometimes get left behind (always return true, as resource may have been successfully been cleaned up)
 
     # Detach any policies from the EKS node role
-    aws --region "$REGION" iam list-attached-role-policies --role-name "eks-${CLUSTERNAME}-node" --output json | jq '.AttachedPolicies[].PolicyArn' | xargs -r -L1 aws --no-cli-pager iam detach-role-policy --role-name "eks-${CLUSTERNAME}-node" --policy-arn || true
+    aws --region "$REGION" iam list-attached-role-policies --role-name "eks-${CLUSTERNAME}-node" --output json | jq '.AttachedPolicies[].PolicyArn' | xargs -tr -L1 aws --no-cli-pager iam detach-role-policy --role-name "eks-${CLUSTERNAME}-node" --policy-arn || true
 
     # Delete EKS IAM roles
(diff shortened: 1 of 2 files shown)


src/qa-test-eks.sh
```diff
@@ -146,4 +170,7 @@ if [ "$ACTION" = "destroy-shared" ]; then
     
     # Cleanup
     terragrunt destroy-all --terragrunt-working-dir "$WORKDIR" --terragrunt-source-update --terragrunt-non-interactive -input=false -auto-approve || true
```
avnes, line 172: Should terragrunt really always succeed? What if it actually fails due to a timeout or a lock? Then there will be resources left behind?

avnes (changes requested):

src/qa-test-eks.sh
```diff
@@ -1,15 +1,31 @@
...
 
-function extra_cleanup {
+function cleanup_cluster {
 
     # Remove specific resources that sometimes get left behind (always return true, as resource may have been successfully been cleaned up)
-    aws --region "$REGION" iam list-roles --output json | jq -r --arg ROLEPREFIX "eks-${CLUSTERNAME}-" '.Roles[] | select( .RoleName | contains($ROLEPREFIX, "Velero") ) | .RoleName' | xargs -r -L1 aws --region "$REGION" iam delete-role --role-name || true
-    aws --region "$REGION" ec2 describe-network-interfaces --filters "Name=group-name,Values=eks-${CLUSTERNAME}-node" --query "NetworkInterfaces[].NetworkInterfaceId" --output text | xargs -r -L1 aws --region "$REGION" ec2 delete-network-interface --network-interface-id || true
+
+    # Detach any policies from the EKS node role
+    aws --region "$REGION" iam list-attached-role-policies --role-name "eks-${CLUSTERNAME}-node" --output json | jq '.AttachedPolicies[].PolicyArn' | xargs -tr -L1 aws --no-cli-pager iam detach-role-policy --role-name "eks-${CLUSTERNAME}-node" --policy-arn || true
```
rifisdfds, line 13: Shouldn't we be looping through all the eks-clustername roles it finds and detaching policies and instance profiles from them, rather than being specific to just policies on eks-node? It'd potentially catch more

avnes (approved):

avnes (approved):

PR #227, opened 2021-06-11
Traefik pod topology spread

abstrask:
Currently, we use anti-affinity to ensure Traefik instances were distributed across our different availability zones. This mean:

- We can not scale the number of Traefik instances beyond the number of AZs (currently 3)
- If an AZ goes down, the Traefik instance going down with it, cannot be rescheduled to other AZs, decreasing the capacity of the Traefik service

This PR replaces pod anti-affinity with the more flexible "[Pod topology spread constraints](https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/)". With this, you can use a topology to distribute pods on different AZs, while still allowing some deviation, which in turn allows all the desired instanced to be schedule-able.

On top of this, a second constraint will prevent multiple instances to be scheduled on the same node, but allowing a `max_skew` of 1, to allow for rolling updates on minimal clusters.

This PR includes a significant version bump for the Kubernetes provider.

### DIFF
--- a/_sub/compute/k8s-traefik-v2/main.tf
+++ b/_sub/compute/k8s-traefik-v2/main.tf
@@ -127,7 +127,7 @@ resource "kubernetes_deployment" "traefik" {
           name  = "traefik"
 
           resources {
-            requests {
+            requests = {
               cpu    = var.request_cpu
               memory = var.request_memory
             }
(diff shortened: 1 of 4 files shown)


wcarlsen (approved):
LGTM!

dfds-pwes (approved):
Looks good to me.

PR #228, opened 2021-06-14
add extra call to cluster cleanup in qa script

rifisdfds:
- add extra call to cleanup_roles with clustername- prefix


dfds-pwes (approved):
LGTM

PR #229, opened 2021-06-14
Update provider versions for the following providers.

dfds-pwes:
hashicorp/aws
hashicorp/kubernetes
hashicorp/helm

I have also removed the version constraints on individual provider blocks because I think the intent is that the version lock is applied in the required_providers section.  There were lots of these for AWS and it just made the process of bumping the version up in future more complicated.

I have tested the above changes in a sandbox both as a fresh deploy and also applying on top of an existing cluster (to map the behaviour we'll see when we merge this and it executes against the production cluster).  I've not seen any issues at all.

### DIFF
--- a/_sub/examples/route53-zone/main.tf
+++ b/_sub/examples/route53-zone/main.tf
@@ -1,5 +1,4 @@
 provider "aws" {
-  version = "~> 2.43"
   region  = var.aws_region
 }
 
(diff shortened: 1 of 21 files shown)


wcarlsen (approved):
Great and good you fixed the versions in regular provider blocks as well

abstrask (approved):

PR #231, opened 2021-06-15
Atlantis bump resource limits

abstrask:
100m CPU and 1Gi seems to constraint for the upper limits of Atlantis.

Bump to 1 CPU and 2 Gi memory (limit).


dfds-pwes (approved):
LGTM

PR #232, opened 2021-06-16
Add support for EKS 1.20

avnes:
With a tactical fix for clusterrole when upgrading to coredns 1.8.3


_sub/compute/eks-addons/main.tf
```diff
@@ -23,6 +23,21 @@ resource "null_resource" "coredns" {
   depends_on = [null_resource.kubeproxy, local.coredns_version]
 }
 
+# For future EKS upgrades, check the upgrade guide for changes to system:coredns clusterrole,
+# and reflect those changes in the YAML file below.
+resource "null_resource" "coredns_clusterrole" {
```
wcarlsen, line 28: This should not be a null resource but regular kubernetes provider resource, which should be possible with clusterroles or a kubectl resource.

wcarlsen (changes requested):
Please only use null resources when official support is available https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs/resources/cluster_role

wcarlsen (approved):
I still don't like the null resourdce clusterrole, but I guess we cannot manage this since CoreDNS is "AWS managing" it. LOL!

abstrask:
#235 seems to negate the need the TF changes. If you agree, maybe we can roll off everything except adding 1.20 to QA? @avnes

abstrask (changes requested):
#235 seems to negate the need the TF changes. If you agree, maybe we can roll off everything except adding 1.20 to QA?

avnes:
> #235 seems to negate the need the TF changes. If you agree, maybe we can roll off everything except adding 1.20 to QA?

I am currently testing #235, and assuming it is successfull I will cancel this PR, and rather add a new PR for QA only. That way we won't pollute git history with reverts.

avnes:
I'll close this PR and open a new one for QA pipeline for EKS 1.20

PR #238, opened 2021-06-24
Avoid rapid ALB target group deregistation

wcarlsen:
Setting the deregistration delay back to its default 300s (https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lb_target_group#deregistration_delay) in alignment with this postmortem https://github.com/dfds/postmortems/blob/master/PM2021-005%20-%20All%20services%20returning%20503%20in%20Kubernetes.md#contributing-factors

### DIFF
--- a/_sub/compute/eks-alb-auth/main.tf
+++ b/_sub/compute/eks-alb-auth/main.tf
@@ -19,7 +19,7 @@ resource "aws_lb_target_group" "traefik_auth" {
   port                 = var.target_http_port
   protocol             = "HTTP"
   vpc_id               = var.vpc_id
-  deregistration_delay = 0
+  deregistration_delay = 300
 
   health_check {
     path     = var.health_check_path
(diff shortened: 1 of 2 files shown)


abstrask:
The default is 300s, so we could also just remove the property, as it was before we explicitly lowered it to 0.

It's a matter of preference, I guess. Do we want to be explicit about it, or rely on AWS' "sane defaults" as we did prior to the change?

I can swing either way. Just curious what your thoughts are.

rifisdfds (approved):

rifisdfds:
I'm happy for us to explicitly set the value, at least we know it's not going to change outside of our control

wcarlsen:
I had the same thoughts @abstrask and went for setting it explicit and then we can remove it later, when we learn more about the behavior.

abstrask:
👍

avnes (approved):

PR #242, opened 2021-06-30
Remove traefik okta

abstrask:
The Traefik v2/Okta instance was deployed specifically to support forward-auth for Backstage. However Backstage is using our old ingress controller and no other services are using this Traefik instance.

Querying our LB access logs also confirms that we only receive garbage on this load balancer, directly on its IP address.

```sql
SELECT request_url, Count(*) AS Count
FROM alb_logs
WHERE elb = 'app/hellman-traefik-okta/3ac5266888aa60a0'
GROUP BY request_url
ORDER BY Count DESC
LIMIT 100;
```

(https://eu-west-1.console.aws.amazon.com/athena/home?region=eu-west-1#query/history/64b14fae-5d22-4799-9504-4fde3d39b65b)

This PR removes the Traefik v2/Okta from the repo entirely. The new Traefik v2 deployment we are working on a different branch, uses a different way of deploying Traefik.


avnes (dismissed):

avnes:
@raras-dfds This PR looks fine, but you also need to make a PR on eks-pipeline because it references this module here: https://github.com/dfds/eks-pipeline/blob/a1def543919fb398dd998f0b9ca3e8c4ae627076/oxygen-account/eu-west-1/k8s-hellman/services/terragrunt.hcl

abstrask:
Forgot to commit a minor additional change. Can you take a second look @avnes ?

avnes (approved):

PR #247, opened 2021-07-06
Enable allow volume expansion on storage classes

dfds-pwes:
Removal of null resource used to set csi-gp2 as default storage class.  Now handled using the annotations in metadata.
Set allowVolumeExpansion to true for the csi-gp2 storage class.  Note that this only works for new clusters unless you taint the existing instance of the storage class.

### DIFF
--- a/_sub/compute/helm-ebs-csi-driver/main.tf
+++ b/_sub/compute/helm-ebs-csi-driver/main.tf
@@ -100,31 +100,34 @@ resource "helm_release" "aws-ebs-csi-driver" {
 resource "kubernetes_storage_class" "csi-gp2" {
   metadata {
     name = "csi-gp2"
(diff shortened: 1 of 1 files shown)


avnes (approved):
Are there any plans to do the migration on excisting clusters? If 'yes', it might be wort adding a CHANGELOG.md file to this repo with instructions on how to do that when applying the new release and migrating.

dfds-pwes:
@avnes The only real persistent cluster we currently have is Hellman and that one I will patch manually.  Your sandbox will also need updating manually, but given that sandboxes are often destroyed and recreated then the likelihood is that over time you'll end up with the changes anyway.  Similar we have the QA clusters, but again those get created and then destroyed each time the QA cycle runs.

PR #249, opened 2021-07-06
Add traefik v2 through Flux

avnes:
- There was a race condition when the fallback ingressroute was created before the traefik v2 resources that essentially blocked traefik from being deployed in the first place. Flux CD could simply not reconcile. The workaround to this was to use a null_resource to wait for the ingressroute created by Helm to be created first.

In order to test from sandbox 4 steps are needed

1. In services/terragrunt.hcl:
  traefik_flux_with_fallback = true
  traefik_flux_github_owner = "dfds"
  traefik_flux_repo_name = "platform-manifests-qa"

2. In https://github.com/dfds/platform-manifests-qa/ copy the apps/volantis/traefik folder over to your apps folder
3. In https://github.com/dfds/platform-manifests-qa/ copy the clusters/volantis/platform-apps.yaml over to your clusters folder
4. In https://github.com/dfds/platform-manifests-qa/ edit your cluster's apps.yaml and add a dependency to  platform-apps-sources as seen here: https://github.com/dfds/platform-manifests-qa/blob/main/clusters/volantis/apps.yaml#L10


_sub/compute/k8s-traefik-flux/main.tf
```diff
@@ -0,0 +1,18 @@
...
+
+resource "github_repository_file" "traefik_patch" {
+  repository = var.repo_name
+  branch     = local.repo_branch
+  file       = "${local.base_repo_path}/patch.yaml"
+  content    = jsonencode(local.patch)
+}
+
+resource "kubectl_manifest" "traefik_fallback" {
+    count       = var.fallback ? 1 : 0
```
wcarlsen, line 16: Could this variable be named fallback_enabled or something to clarify it's purpose better.

_sub/compute/k8s-traefik-flux/vars.tf
```diff
@@ -0,0 +1,89 @@
...
+variable "fallback_ingressroute_name" {
+  type        = string
+  description = "The name for the ingressroute used for fallback"
+  default     = "traefik-fallback-to-v1-ingress"
+}
+
+variable "fallback_ingressroute_priority" {
+  type        = number
+  description = "IngressRoute priority. Should be a low number, but preferably not lower than 2"
+  default     = 2
```
wcarlsen, line 70: This could also be 1, but I generally like the priority of 2.

_sub/compute/k8s-traefik-flux/vars.tf
```diff
@@ -0,0 +1,89 @@
...
+}
+
+variable "fallback_svc_namespace" {
+  type        = string
+  description = "The service used for fallback ingress is stored in which namespace"
+  default     = "kube-system"
```
wcarlsen, line 76: I don't think defaults makes any sense here, because you are assuming our implementation in the sub module.

_sub/compute/k8s-traefik-flux/vars.tf
```diff
@@ -0,0 +1,89 @@
...
+}
+
+variable "fallback_svc_name" {
+  type        = string
+  description = "The service name used for fallback ingress"
+  default     = "traefik"
```
wcarlsen, line 82: I don't think defaults makes any sense here, because you are assuming our implementation in the sub module.

_sub/compute/k8s-traefik-flux/vars.tf
```diff
@@ -0,0 +1,89 @@
...
+}
+
+variable "fallback_svc_port" {
+  type        = number
+  description = "The service port used for fallback ingress"
+  default     = 80
```
wcarlsen, line 88: I don't think defaults makes any sense here, because you are assuming our implementation in the sub module.

wcarlsen (dismissed):
LGTM! But if you removed the defaults from the fallback variables it would be even better.

wcarlsen (approved):
LGTM!

PR #253, opened 2021-07-08
Bump EKS QA pipeline image (Terraform 1.x)

abstrask:


avnes (approved):
I tried this in my sandbox too, and I did not see any problems with it.

PR #259, opened 2021-07-09
More deprecation fixes

abstrask:
Primarily moving provider version locks into the `required_providers` block in `versions.tf`, but also removing some old comments.


wcarlsen (approved):
LGTM!

PR #262, opened 2021-07-12
Fix for TF linter being so helpful, it removed an curly end-brace

abstrask:
Bloody TF linter!....


dfds-pwes (approved):
LGTM - so many curly braces!!

SEQUOIIA:
CURRRRRLLLLY

PR #265, opened 2021-07-21
Traefikv2 ALB config

dfds-pwes:
- Adds flag to allow test ALB's to be deployed with routing to Traefik v2 Node Ports.
- Adds version lock option for Traefik v2 Helm Chart.
-

### DIFF
--- a/compute/k8s-services/main.tf
+++ b/compute/k8s-services/main.tf
@@ -248,6 +248,46 @@ module "traefik_alb_anon_dns_core_alias" {
 
 
 # --------------------------------------------------
+# Traefik v2: For validation and testing
+# --------------------------------------------------
+
+module "traefikv2_alb_anon" {
+  source                = "../../_sub/compute/eks-alb"
+  deploy                = var.traefikv2_test_alb_deploy
+  name                  = "${var.eks_cluster_name}-traefikv2-alb"
+  cluster_name          = var.eks_cluster_name
+  vpc_id                = data.aws_eks_cluster.eks.vpc_config[0].vpc_id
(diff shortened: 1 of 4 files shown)


compute/k8s-services/main.tf
```diff
@@ -247,6 +248,46 @@ module "traefik_alb_anon_dns_core_alias" {
 }
 
 
+# --------------------------------------------------
+# Traefik v2: For validation and testing
+# --------------------------------------------------
+
+module "traefikv2_alb_anon" {
+  source                = "../../_sub/compute/eks-alb"
+  deploy                = var.traefikv2_test_alb_deploy
```
rifisdfds, line 257: Do we want the word "test" in the variable name?

dfds-pwes: I'm honestly not sure.

To give a bit of background the intention is that this variable will deploy some ALB's that will only be used for testing the new Traefik v2 implementation.  Once it's tested and confirmed as working then they will be torn down (by changing the variable to false) and then the existing ALB's will be updated.

We could perhaps use traefikv2_migration_alb_deploy instead possibly but i'm not convinced the name conveys their real purpose.

rifisdfds: Ok, then I'm happy with whatever you go with

rifisdfds (approved):

compute/k8s-services/main.tf
dfds-pwes: We will stick with test for now, but only on the basis that once we're done with the testing phase then the code will most likely be stripped from the module anyway.  Rasmus had originally said to me about possibly doing the code outside of infrastructure-modules, but doing so would have been much more trouble because of the dependencies.

PR #267, opened 2021-07-28
Add KubeletTooManyPods warning alert to null receiver

rifisdfds:
This PR sends KubeletTooManyPods to the null receiver to silence alerts in the hellman-alerting channel. We are currently receiving these due to the 58 pod limit (due to ENI) on our eks node instance sizes, of which 2 nodes have triggered. We should be covered by the Node Capacity alert from Grafana if we have too many pods across the entire cluster to withstand an AZ failure so this warning alert is currently noise


dfds-pwes (approved):
Looks good to me.

PR #268, opened 2021-07-29
Fix qa destroy shared

rifisdfds:
This PR fixes the issue with QA being unable to remove shared roles (i.e Velero) due to a missing region variable by passing the region through the pipeline to the shell script


dfds-pwes (approved):
LGTM

PR #270, opened 2021-08-02
Add support for EKS 1.21 to EKS add-ons and to QA pipeline

avnes:


dfds-pwes (commented):
Are we not running behind on the version of the VPC CNI add-on we're using?

The linked article suggests we should be on 1.9.x?

https://docs.aws.amazon.com/eks/latest/userguide/managing-vpc-cni.html

avnes:
> Are we not running behind on the version of the VPC CNI add-on we're using?
> 
> The linked article suggests we should be on 1.9.x?
> 
> https://docs.aws.amazon.com/eks/latest/userguide/managing-vpc-cni.html

Yes, but the command `aws eks describe-addon-versions     --addon-name vpc-cni     --kubernetes-version 1.21     --query "addons[].addonVersions[].[addonVersion, compatibilities[].defaultVersion]"     --output text --region eu-west-1` says that the 1.7.5 version is still the default version.

Results:
```
v1.9.0-eksbuild.1
False
v1.8.0-eksbuild.1
False
v1.7.10-eksbuild.1
False
v1.7.9-eksbuild.2
False
v1.7.6-eksbuild.2
False
v1.7.5-eksbuild.2
True
v1.6.3-eksbuild.2
False
```

The other concern is that we can't directly upgrade from 1.7.5 to 1.9.0, we will have to go through a 1.8.x upgrade first, so it is probably better to raise that as a separate enhancement.

Reference:
`Updating VPC-CNI can only go up or down 1 minor version at a time`

dfds-pwes (approved):
LGTM

PR #273, opened 2021-08-05
Exclude all namespaced alerts not matching target namespaces feature

wcarlsen:
We can now get rid of and stop maintaining that long list of unwanted namespace alerts.

See example config on [alertmanager's github page](https://github.com/prometheus/alertmanager#example) to understand more in dept how it works.


wcarlsen:
TL;DR;
Basically all alerts with a label "namespace" will be sent to the null receiver (aka black hole of alerts), but if it matches one of the target namespaces it will be sent to our slack channel receiver (general). Hence we have now hacked a negative matcher by using sub-routes.

avnes (approved):

PR #274, opened 2021-08-06
Fix Traefik install and config race conditions

avnes:
It breaks the platform apps into multiple pieces to avoid using one parent app with sub folders (like apps/cluster/app-name(s)), because in that scenario one app with problems can kill reconciliation for all other apps under the same structure. 

Now there is one kustomization called platform-apps-sources which imports Flux sources from https://github.com/dfds/platform-apps/tree/main/sources. 

For each app in platform-apps, there will be a **appname**-helm (for installing and patching the helm chart) and a **appname**-config kustomization for changes that needs to go in AFTER the CRDs are in place. The contents in **appname**-helm and **appname**-config is created using Terraform.

So for traefik it will look like this:

```
flux get kustomizations
NAME                        	READY
apps                        	True
flux-system                 	True
infrastructure              	True
platform-apps-sources       	True
platform-apps-traefik-config	True
platform-apps-traefik-helm  	True
```

### DIFF
--- a/_sub/compute/k8s-traefik-flux/dependencies.tf
+++ b/_sub/compute/k8s-traefik-flux/dependencies.tf
@@ -5,22 +5,73 @@ data "github_repository" "main" {
 locals {
   default_repo_branch = data.github_repository.main.default_branch
   repo_branch         = length(var.repo_branch) > 0 ? var.repo_branch : local.default_repo_branch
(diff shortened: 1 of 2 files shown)


wcarlsen (approved):
LGTM!

dfds-pwes (approved):
Looks fantastic.  Whoever designed the folder structure is a hero and should be commended!

PR #278, opened 2021-08-11
Add support for passing additional arguments to the Traefik Helm chart 

avnes:
This was made because Traefik Helm chart 10.1.2 introduces a breaking change that denies creating IngressRoute to Service mappings across namespaces. That is good from a security point of view, but it is pretty bad during our migration phase, because we need to fallback to Traefik v1 in another namespace than the fallback ingress-route.

On the implementation side, the service config should for now use the folling settings to allow cross-namespace configuration:

```
traefik_flux_helm_chart_version = "10.1.2"
  traefik_flux_additional_args    = [
    "--metrics.prometheus",
    "--providers.kubernetescrd.allowCrossNamespace=true"
  ]
```

### DIFF
--- a/_sub/compute/k8s-traefik-flux/dependencies.tf
+++ b/_sub/compute/k8s-traefik-flux/dependencies.tf
@@ -93,6 +93,7 @@ locals {
             "nodePort" = var.admin_nodeport
           }
         }
+        "additionalArguments" = var.additional_args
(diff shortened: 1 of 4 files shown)


wcarlsen (approved):
LGTM!

dfds-pwes (approved):
LGTM

PR #279, opened 2021-08-12
Add dependency on Helm install of Atlatis to ensure the NS is there b…

avnes:
…efore saving a k8s secret to the NS

Relates to https://github.com/dfds/cloudplatform/issues/388


dfds-pwes (approved):
LGTM.

PR #290, opened 2021-08-31
Feature/cloudplatform 385

avnes:
- Reconfigure Atlantis to use Basic Authentication through Traefik 2 IngressRoute and Middleware.
- During cleanup in QA, deattach roles (prefixed with eks-$CLUSTER_NAME) from instance profiles and delete the instance profile afterwards in order to avoid errors like: "An error occurred (DeleteConflict) when calling the DeleteRole operation: Cannot delete entity, must remove roles from instance profile first."


abstrask (approved):
LGTM, but I haven't tried it out in sandbox cluster.

PR #294, opened 2021-09-02
Fix problem with updating branches for flux-system

avnes:
Refactor to use data.github_branch to find branch instead of data.github_repository.main.default_branch

Issue: https://github.com/dfds/cloudplatform/issues/348

Tested in QA: https://dev.azure.com/dfds/CloudEngineering/_build/results?buildId=395428&view=results

Also tested in sandbox.

### DIFF
--- a/_sub/compute/k8s-fluxcd/dependencies.tf
+++ b/_sub/compute/k8s-fluxcd/dependencies.tf
@@ -44,14 +44,17 @@ data "github_repository" "main" {
   full_name = "${var.github_owner}/${var.repo_name}"
 }
 
+data "github_branch" "flux_branch" {
+  repository = var.repo_name
+  branch     = var.repo_name
+}
+
 # --------------------------------------------------
 # Monitoring
 # --------------------------------------------------
 
 locals {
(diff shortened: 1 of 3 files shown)


rifisdfds (approved):

ZaradarDFDS (commented):
you seem to have changed this logic: 

length(var.repo_branch) > 0 ? var.repo_branch : local.default_repo_branch

To this:

branch = var.repo_name

Doesnt this indicate that your no longer falling back to the default or is there always a repo name, thus it is no longer required?

avnes:
> you seem to have changed this logic:
> 
> length(var.repo_branch) > 0 ? var.repo_branch : local.default_repo_branch
> 
> To this:
> 
> branch = var.repo_name
> 
> Doesnt this indicate that your no longer falling back to the default or is there always a repo name, thus it is no longer required?

@ZaradarDFDS it would seem so, but the var.repo_name has a default value of "main", which happens to be the name for the default branch for the two repos used by flux in our setup. Hence the logic remains the same.

PR #331, opened 2021-10-01
Cleanup redundant null logic

abstrask:
When applying Helm charts, we have both been setting the `chart_version` variable to `null` by default, and using this redundant logic:

```hcl
version          = var.chart_version != null ? var.chart_version : null
```

This PR cleans up the redundancy, making for more readable code.


wcarlsen (approved):
LGTM!

PR #332, opened 2021-10-01
Bump required Terraform version to 1.0 for all modules

abstrask:
We no longer validate any of our modules with Terraform versions older than 1.0, and thus the required Terraform version is now effectively 1.0.


wcarlsen (commented):
LGTM!

wcarlsen (approved):

PR #337, opened 2021-10-05
Fixing one QA issue and attempting to fix a second QA issue. Thirdly disable app reg

avnes:
This PR contains to changes to QA:

1. Adding a 'sleep 900' before deleting ENI resources. This was mentioned as a workaround in a issue related to the AWS CLI when deleting ENI.
2. The QA pipeline was out of sync with the crossplane releases used in production. So this has now been aligned.
3. Due to a bug in AzureAD provider, I have disabled the app registration in QA for now.


wcarlsen (approved):
I'm so much not fan of the sleep 900, but I guess there is no way around it. Please warn the others that running QA during work hours is almost not allowed anymore.

PR #357, opened 2021-11-10
Crossplane - Use UXP release

SEQUOIIA:
Seems to work on a "fresh" cluster without any previous Crossplane installations.

Currently testing the following scenarios:
- [x] (Yes) Does it work on a fresh install
- [x] (Surprisingly yes?) Does it work if you go from the open-source version of Crossplane to this
  - [x] (Yes) Do you keep your resources?
  - [x] (Tested Bucket & RDSInstance, everything was kept intact) If not, are they orphaned at least?

### DIFF
--- a/_sub/compute/helm-crossplane/main.tf
+++ b/_sub/compute/helm-crossplane/main.tf
@@ -1,12 +1,12 @@
 resource "helm_release" "crossplane" {
   name          = var.release_name
-  chart         = "crossplane"
-  repository    = "https://charts.crossplane.io/stable"
+  chart         = "universal-crossplane"
+  repository    = "https://charts.upbound.io/stable"
   version       = var.chart_version
   namespace     = var.namespace
   recreate_pods = var.recreate_pods
   force_update  = var.force_update
-
+  devel         = var.devel
   values = [
     templatefile("${path.module}/values/values.yaml", {
       crossplane_metrics_enabled = var.crossplane_metrics_enabled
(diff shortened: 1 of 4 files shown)


SEQUOIIA:
I would love for one of you, @rifisdfds or @samidbb if you could test the scenarios mentioned in the PR description in a sandbox cluster. I got it working as intended, but could just be my setup 😅

So:

1. Apply this infrastructure-modules with this branch to your sandbox cluster
2. Test the scenarios from the PR descriptions

Let me know if you have any questions 😀

rifisdfds:
@SEQUOIIA I've tried an upgrade and with the deletion of crossplane-system namespace and replacement with upbound-system, the secrets that the compositions place in there get removed, and the synced status of a deployed dynamic database claim becomes false, presumably as a result.

Fortunately, as we don't have any claims deployed in production we should get away with it but I'll raise an issue for us to edit our compositions to use upbound-system instead or perhaps patch in the namespace (though this may break static claims and we may not be able to)

Raw (non-composition) resources seemed to come through without any issue

wcarlsen (approved):
LGTM!

rifisdfds (approved):

samidbb:
unfortunately I have overseen this issue because of the all the spam I have got from github notifications.. otherwise I would have thought about change of namespace as an issue........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................

PR #361, opened 2021-11-17
fix known_hosts value due to github rotation

rifisdfds:
This PR changes the known_hosts value for Github as they have removed support for their old DSA key. Value now uses ECSDA


dfds-pwes (approved):
LGTM

PR #368, opened 2021-12-02
Remove tfsec alerts that are either:

avnes:
- In irrelevant modules.
- False positives
- Easy to fix and without impact.


wcarlsen (approved):
Nice!

jdsmithit (approved):

PR #381, opened 2022-01-12
Use local variable in main.tf

cristian-dfds:
The engine_family variable is declared locally, but it is used as an input variable so  the pipeline fails when it runs because it doesn't find a value. This PR fixes the bug.


avnes (approved):
LGTM

PR #383, opened 2022-01-14
Create before destroy parameter group for RDS DB

cristian-dfds:


_sub/database/postgres/main.tf
```diff
@@ -36,6 +36,10 @@ resource "aws_db_parameter_group" "dbparams" {
     value = "1"
   }
 
+  lifecycle {
+    create_before_destroy = true
```
avnes, line 40: One concern: "Destroy provisioners of this resource will not run if create_before_destroy is set to true" 

Ideally we would want to make this lifecycle change to be optional, and it should rather be controlled through parameters, like:

`create_before_destroy = var.create_before_destroy`

In _sub/database/postgres/vars.tf then you also define this create_before_destroy bool variable with a default value of true

avnes (changes requested):

_sub/database/postgres/main.tf
cristian-dfds: The security group defined above the parameter group has the same lifecycle defined. Should that also reference the 
`var.create_before_destroy` parameter I will create.

avnes: Since you did not add that to the security group, I would say it would be scope creep for now. The change looks good to me as it is now.

avnes (approved):

PR #396, opened 2022-01-27
Feature/cloudplatform 580

avnes:
Fix a bug in the Flux sub module where the wrong variable is passed over to a data object.


wiredworm (approved):
LGTM

PR #399, opened 2022-01-31
Flux migration: Podinfo and fluend-cloudwatch

avnes:


_sub/monitoring/fluentd-cloudwatch/vars.tf
```diff
@@ -0,0 +1,32 @@
...
+variable "github_owner" {
+  type        = string
+  description = ""
+}
+
+variable "aws_region" {
+  type = string
+}
+
+variable "aws_workload_account_id" {
```
wiredworm, line 31: Are you able to have a variable in Terraform without a defined typed?

wiredworm (dismissed):
Looks good.  Just a couple of comments:

1. There are some missing descriptions on the variables.  Not sure if that needs addressing, as otherwise Rasmus might get mad!  ;-)
2. The known hosts bit relating to GitHub is something that I think the later versions of the provider are meant to take care of.  I seem to recall reading about that when me and Richard had to update the module, but at that time we were stuck with an old provider.

Otherwise it all looks good.

Pete

_sub/monitoring/fluentd-cloudwatch/vars.tf
avnes: The reason why is that it was copied like that from another sub module (kiam)

wiredworm (approved):
LGTM

PR #400, opened 2022-01-31
Add ability to override which account to store the fluentd logs.

avnes:
Add support to override log file retention policy

Signed-off-by: Audun Nes <aunes@dfds.com>


wiredworm (approved):
LGTM

PR #417, opened 2022-02-11
Replace KIAM with IRSA for Velero

avnes:
This PR will remove KIAM for Velero, and instead use IAM Role for Service Account.


abstrask (approved):
Suggest renaming `oidc_provider_server_id` to just `oidc_provider_id`. Otherwise, looks good.

PR #419, opened 2022-02-14
Feature/add crossplane kubernetes provider

rifisdfds:
This PR adds the crossplane kubernetes provider if requested


samidbb (approved):
I suggest spawning another issue for the cleanup part

PR #424, opened 2022-02-17
OIDC provider isolation

wiredworm:
Migrates the OIDC Provider into it's own module rather than having it included as part of the CSI EBS Driver module.

### DIFF
--- a/_sub/compute/helm-ebs-csi-driver/vars.tf
+++ b/_sub/compute/helm-ebs-csi-driver/vars.tf
@@ -26,9 +26,3 @@ variable "csi_ebs_serviceaccount_name" {
   description = "The name of the Service Account used by the CSI EBS Driver."
   default     = "ebs-csi-controller-sa"
 }
-
-variable "csi_ebs_serviceaccount_namespace" {
-  type        = string
-  description = "The namespace  where the Service Account used by the CSI EBS Driver is located."
-  default     = "kube-system"
-}
--- a/_sub/compute/helm-ebs-csi-driver/vars.tf
+++ b/_sub/compute/helm-ebs-csi-driver/vars.tf
@@ -21,7 +21,14 @@ variable "cluster_name" {
   description = "The cluster name"
 }
 
-variable "kiam_server_role_arn" {
+variable "csi_ebs_serviceaccount_name" {
   type        = string
-  description = "The role or entity to provide trust for when creating roles to use with annotations in kubernetes"
+  description = "The name of the Service Account used by the CSI EBS Driver."
+  default     = "ebs-csi-controller-sa"
+}
+
+variable "csi_ebs_serviceaccount_namespace" {
+  type        = string
+  description = "The namespace  where the Service Account used by the CSI EBS Driver is located."
+  default     = "kube-system"
 }
--- a/compute/k8s-services/main.tf
+++ b/compute/k8s-services/main.tf
@@ -81,6 +81,14 @@ provider "azuread" {
 
 }
 
+# --------------------------------------------------
+# AWS IAM Open ID Connect Provider
+# --------------------------------------------------
+
+module "aws_iam_oidc_provider" {
+  source                          = "../../_sub/security/iam-oidc-provider"
+  eks_openid_connect_provider_url = data.aws_eks_cluster.eks.identity[0].oidc[0].issuer
+}
 
 # --------------------------------------------------
 # AWS EBS CSI Driver (Helm Chart Installation)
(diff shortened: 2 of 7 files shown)


avnes (commented):
This certainly seem to cater for the use case I have, so I am happy to approve this PR.

I do have one question though:
- Should we hardcode the audience to a list containing sts.amazonaws.com? Or should it be through a variable of type list that defaults to containing sts.amazonaws.com?

wiredworm:
My understanding is that the provider we define in IAM is only to allow IAM to communicate with the EKS Cluster and on that basis the audience should always be sts.amazonaws.com.

I did some digging through various documents and they all stipulated that value.  It's possible it could change in the future I guess but that's not assured so we'd be kind of second guessing what might happen.  On that basis I think we hard-code it for now, but just keep in mind that it could be passed in as a variable should it be required at a later date.

_sub/compute/helm-ebs-csi-driver/vars.tf
```diff
@@ -21,7 +21,14 @@ variable "cluster_name" {
   description = "The cluster name"
 }
 
-variable "kiam_server_role_arn" {
+variable "csi_ebs_serviceaccount_name" {
```
wcarlsen, line 24: Why are these variables, when they are hardcoded and not even used nor exposed in k8s-services. Could they be either locals or be actual referenced in a set block for the helm chart?

_sub/compute/helm-ebs-csi-driver/vars.tf
```diff
@@ -21,7 +21,14 @@ variable "cluster_name" {
...
   type        = string
-  description = "The role or entity to provide trust for when creating roles to use with annotations in kubernetes"
+  description = "The name of the Service Account used by the CSI EBS Driver."
+  default     = "ebs-csi-controller-sa"
+}
+
+variable "csi_ebs_serviceaccount_namespace" {
+  type        = string
+  description = "The namespace  where the Service Account used by the CSI EBS Driver is located."
+  default     = "kube-system"
```
wcarlsen, line 33: Why are these variables, when they are hardcoded and not even used nor exposed in k8s-services. Could they be either locals or be actual referenced in a set block for the helm chart?

wcarlsen (changes requested):
Either use the variables csi_ebs_serviceaccount_name and csi_ebs_serviceaccount_namespace  for other than hardcoding stuff (i.e. use locals) or use a set block for the variables :+1:

wcarlsen:
@wiredworm am I correct that the module wouldn't work if I actually set those variables?

wiredworm:
@wcarlsen You are probably correct and i've maybe over-thought things a bit too much.  I'll revert those changes back so we just have the hard-coded values.

wcarlsen:
@wiredworm with that approach you risk changes in the helm chart like serviceaccount name to break the module

samidbb (dismissed):

compute/k8s-services/main.tf
```diff
@@ -523,7 +530,7 @@ module "crossplane" {
   crossplane_view_service_accounts  = var.crossplane_view_service_accounts
   crossplane_metrics_enabled        = var.crossplane_metrics_enabled
   crossplane_aws_iam_role_name      = var.crossplane_aws_iam_role_name
-  eks_openid_connect_provider_url   = var.eks_openid_connect_provider_url
```
rifisdfds, line 526: You should remove the var from k8s-services vars.tf with no longer referring to it as it is redundant

rifisdfds (approved):

compute/k8s-services/main.tf
wiredworm: Strangely the code shown above suggest i've entirely removed that section.

The line itself is still valid but instead of referencing a var it simply pulls it directly from the datasourcenow.

rifisdfds: you have replaced var.eks_openid_connect_provider_url with data.aws_eks_cluster.eks.identity[0].oidc[0].issuer
but you have not deleted the eks_openidconnect_provider_url variable from the vars.tf file as far as i can see at https://github.com/dfds/infrastructure-modules/blob/551aff5f9fc0f29d9b52ce353e754dc4669d5464/_sub/compute/helm-crossplane/vars.tf#L70 which is no longer required since you removed the use of it so it is sat there unused/redundant

wiredworm:
I've now modified things so that the namespace for the service-account is hard-coded to kube-system (as there is no means to configure this via the Helm Chart).  The variable for the service account name has a default set and this is now specified in the values used to deploy the helm chart; so it won't matter if no variable is given or not, it will always work.

wcarlsen (approved):
LGTM!

PR #428, opened 2022-02-18
Added separate IAM role for k8s-janitor

SEQUOIIA:


_sub/compute/k8s-blaster-namespace/dependencies.tf
```diff
@@ -0,0 +1,67 @@
+
+data "aws_caller_identity" "workload_account" {
```
wcarlsen, line 2: Many ways to get this either via a variable or in this way. I honestly thinks this is way better, but I also don't really care, since we are using the variable approach all over.

wcarlsen (approved):
LGTM!

PR #429, opened 2022-02-21
Migrate OIDC provider to cluster phase

wiredworm:


wcarlsen:
Also I guess this means a new oidc issuer url and we need to check if anyone has it hardcoded somewhere.

abstrask:
@wcarlsen Didn't we also do that, when it was moved to a different module? I checked yesterday and noticed the OIDC provider was very young (3 days AFAIR). So if it was to break, I would assume it already had.

Interestingly the newly created provider has the same ID, and thus endpoint, as the old one. Wonder if the ID is derived from the the cluster, or cluster name, somehow?

abstrask (dismissed):
ClapClappingGIF

wiredworm:
I believe the URL is created when the cluster is first created, so for long lived clusters we shouldn't see any issue.  And of course for those that aren't long-lived it wouldn't be an issue anyway.  So I think we are safe.

abstrask (dismissed):
HomerSimpsonPigGIF

abstrask (commented):
Sorry to be difficult, @wiredworm, but I don't think we've ever referenced other sub modules from sub modules. 

Until now we've only ever referenced sub modules from "main" modules.

Is that a route we want to take?

wiredworm:
> Is that a route we want to take?

I have no idea.

From a technical perspective it works.  If there was a risk of a circular dependency then i'd probably be concerned, but the iam-oidc-provider contains so little complexity that I don't see it as a problem.

From a functional perspective it's identical to just having the resources directly in eks-cluster.  So I guess we either accept this as the best option, or the other would be to duplicate the code and have it both in the module and also directly in eks-cluster.

I don't mind either option, so if there is a real concern about nesting sub-modules then we accept duplicate code instead.

avnes (dismissed):

wcarlsen (changes requested):
_sub module may not reference each other in the current structure of infrastructure-modules (how ever stupid this may sound). Consider moving references across modules to a main module, so that we can everybody make hard coupling to your module, so it can never be refactored easy again :+1:  or do code duplication.

wiredworm:
Thanks @wcarlsen 

@abstrask It seems that if we want to stick to the design pattern then we must not reference one sub-module from another.  This leaves us with a couple of options:

1. Duplicate the code in the iam-oidc-provider module and include this in the eks-cluster module directly.  It means some code duplication but allows us to contain the OIDC Provider creation with eks-cluster.
2. Move the creation of the OIDC Provider to the main module (eks-ec2) instead and reference the iam-oidc-provider sub-module from there.

The volume of code involved is only around 20 lines and it's not super complex.  Personally I don't have an issue with option 1, particularly not if we attach comments to make it clear as to why the code is duplicated and where the similar code also lives.

abstrask:
@wiredworm I'll suggest we go with option 2, though I think either would suit us just fine. We may end up creating OIDC providers in every capability account(?), so it may be handy to keep it as a sub module. And if we already have a sub module, we might as well use that, instead of duplicating code.

Any objections, @wcarlsen ?

wcarlsen (approved):
LGTM!

PR #430, opened 2022-02-21
Use intended variable for iam policy rather than having it hardcoded

SEQUOIIA:
Doesn't change anything for Hellman. Does make our life a whole lot easier for any cluster not named Hellman (sandbox clusters 😄 )

The variable for Hellman is set at: https://github.com/dfds/eks-pipeline/blob/75e9ed7c9bb25666fae5f6734a54b283aa7a555e/oxygen-account/eu-west-1/k8s-hellman/cluster/terragrunt.hcl#L38


wcarlsen (approved):
LGTM!

PR #432, opened 2022-02-22
maintenance/fluentd kiam migration

SEQUOIIA:
Replaced KIAM with IAM OIDC for fluentd setup.


avnes (approved):
LGTM

SEQUOIIA:
Looking to merge this on Friday where we have a set maintenance window for this in case something doesn't go as planned.

PR #437, opened 2022-02-25
Added region to k8s auth exec command

SEQUOIIA:
After looking at various online sources & usage of "aws eks get-token", it looks like adding the AWS region explictly may fix the issue we see in QA.

For QA logs, see run "442679" in ADO.

Sources:
https://florian.ec/blog/github-actions-awscli-errors/
https://github.com/terraform-aws-modules/terraform-aws-eks/pull/1699/files


wcarlsen (approved):
LGTM

PR #485, opened 2022-05-17
modify validation condition on velero variable to validate default

rifisdfds:
This PR allows the `velero_plugin_for_aws_version` and `velero_plugin_for_csi_version` variables to pass validation if no value is specified and successfully use the default values defined. Typically useful for sandbox deployments but is general good practice

This PR allows the following:

- prevent unnecessary prompts for input values for these variables if `velero_flux_deploy` is set to false by fixing the validation rule at k8s-services level to allow no value
- set default values at k8s-services level rather than an empty string which meant that the default values that were set in the velero module were always overridden

### DIFF
--- a/_sub/storage/velero-flux/vars.tf
+++ b/_sub/storage/velero-flux/vars.tf
@@ -98,7 +98,6 @@ variable "image_tag" {
 
 variable "plugin_for_aws_version" {
   type        = string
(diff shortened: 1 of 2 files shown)


wcarlsen (approved):
LGTM

PR #493, opened 2022-05-31
Replace null_resource and kubectl dep. with native provider resource

abstrask:
As of version 2.10, Kubernetes provider [now supports](https://github.com/hashicorp/terraform-provider-kubernetes/releases/tag/v2.10.0) annotations as a separate Terraform resource, enabling the mangement of annotations that are otherwise not managed by Terraform.

This change removes a `null_resource` and dependency on the `kubectl` binary, in favour of a native provider resource.

Note: I have not been able to test this myself, as I don't currently have an environment where I can do that.


jdsmithit:
This guy looks suspicious..... do you agree? 🥇 FYI @abstrask you are awesome!

PR #504, opened 2022-06-13
Disable TLSv1.0 and TLS1.1 in the AWS load balancers

avnes:


wcarlsen (approved):
LGTM. The NLB is unused and probably should be removed, but I guess this is for another time

PR #527, opened 2022-08-25
Remove KIAM reference from selfservice namespace and associated AWS IAM Role

rifisdfds:


SEQUOIIA (approved):
All selfservice applications runs in my sandbox cluster with this branch without any issues.

PR #528, opened 2022-08-25
Add kiam deploy toggle

rifisdfds:
This PR will add a toggle to allow is to remove the KIAM integration. It will depend upon https://github.com/dfds/infrastructure-modules/pull/527 being merged first


samidbb (approved):
test it working in Sandbox so lgtm

PR #531, opened 2022-09-01
store random grafana password in param store

rifisdfds:
This PR modifies the kube-prometheus-stack module to store the password in the AWS SSM Parameter store. It also increases the length of the randomly generated password


samidbb (approved):
works fine after test in sandbox

PR #532, opened 2022-09-01
mark grafana_admin_password var as sensitive

rifisdfds:
This PR further improves the kube-prometheus-stack module by marking the optional password variable as sensitive to ensure it is not output on a terragrunt plan if a value is provided


samidbb (approved):
working fine

PR #537, opened 2022-09-21
Allow the manage users to be rotated by teams

jdsmithit:
Update IAM SCP to allow the development team to create, remove and rotate the deploy IAM creds,


rifisdfds (changes requested):
I think this PR should also include the removal of the credentials from the parameter store as this policy change allows the SSM parameter store to become out of sync

PR #539, opened 2022-09-27
added toggle for s3 bucket encryption

rifisdfds:
This PR adds a toggle feature to enable/disable server side encryption on an SSE bucket. It defaults to enabled to match the current behaviour, meaning no changes to any inputs are required for existing deployments. You can now, however, optionally disable it by providing `enable_server_side_encryption = false`


matthewhilzinger (approved):
Tested in sandbox.

PR #561, opened 2022-12-02
Remove dependence on Kiam.

emil2k:
Need to be able now to set the Kiam server role in the account contexts, but some of the modules do not work properly when it is not set.

This should fix that.


samidbb (approved):
makes sense

PR #564, opened 2022-12-09
Removed unused variables

emil2k:
Related to https://github.com/dfds/cloudplatform/issues/1152

- Removed unused variables in the EKS pipeline modules, used `tflint` to detect unused variables (related https://github.com/dfds/cloudplatform/issues/1182).
- Some unused variables were introduced intentionally to avoid getting warnings about the setting of undeclared variables. To avoid these warnings I think it is better to refactor the HCLs instead of introducing such unused variables into the modules which were done in a separate PR on the eks-pipeline repo.
- Introduced a validation for a variable, we should probably utilize this more.

### DIFF
--- a/compute/eks-ec2/vars.tf
+++ b/compute/eks-ec2/vars.tf
@@ -1,12 +1,4 @@
 # --------------------------------------------------
-# Terraform
-# --------------------------------------------------
-
-variable "terraform_state_s3_bucket" {
-  type = string
-}
-
-# --------------------------------------------------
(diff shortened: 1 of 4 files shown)


avnes (commented):
@emil2k - LGTM. I am happy to approve when the merge conflicts have been sorted out.

emil2k:
@avnes Thanks, the conflicts are rebased now.

avnes (approved):

PR #567, opened 2022-12-13
Helm exporter addition

Ranixa:


rifisdfds (changes requested):
It needs the Terraform adding to apply the labels/annotation for scraping by Prometheus

avnes:
> It needs the Terraform adding to apply the labels/annotation for scraping by Prometheus

@Ranixa - the problem @rifisdfds refer to is not really in this PR (as far as I can tell), but rather in the HelmRelease Flux manifest at https://github.com/dfds/platform-apps/blob/main/apps/helm-exporter/release.yaml#L22


It tries to scrape metrics using this syntax:

```
  values:
    service:
      labels:
        scrape-service-metrics: "true"
```

According to the helm-chart-exporter Helm values file at https://github.com/sstarcher/helm-exporter/blob/master/helm/values.yaml it doesn't support to create labels under services.

I think the solution here is to update that HelmRelease Flux manifest at https://github.com/dfds/platform-apps/blob/main/apps/helm-exporter/release.yaml#L22 with:

```
values:
  serviceMonitor:
    create: true
    scrapeTimeout: 10s
    additionalLabels:
      release: monitoring
```

_sub/monitoring/helm-exporter/vars.tf
```diff
@@ -0,0 +1,48 @@
...
+
+variable "deploy_name" {
+  type        = string
+  description = "Unique identifier of the deployment, only needs override if deploying multiple instances"
+  default     = "helm-exporter"
+}
+
+variable "namespace" {
+  type        = string
+  description = "The namespace in which to deploy Helm resources"
```
avnes, line 13: Since it is intended that this one is installed in the monitoring namespace, then please add:

`default = "monitoring"`

compute/k8s-services/vars.tf
```diff
@@ -881,6 +881,41 @@ variable "blackbox_exporter_monitoring_targets" {
...
+
+variable "helm_exporter_deploy" {
+  type        = bool
+  description = "Should the helm Exporter be deployed through Flux?"
+  default     = false
+}
+
+variable "helm_exporter_helm_chart_version" {
+  type        = string
+  description = "Helm Chart version to be used to deploy Traefik"
```
avnes, line 896: This ain't Traefik :)

avnes (changes requested):

compute/k8s-services/main.tf
```diff
@@ -582,6 +582,27 @@ module "blackbox_exporter_flux_manifests" {
...
+# --------------------------------------------------
+# helm Exporter
+# --------------------------------------------------
+
+module "helm_exporter_flux_manifests" {
+  source             = "../../_sub/monitoring/helm-exporter"
+  count              = var.helm_exporter_deploy ? 1 : 0
+  cluster_name       = var.eks_cluster_name
+  helm_chart_version = var.helm_exporter_helm_chart_version
+  github_owner       = var.helm_exporter_github_owner
```
avnes, line 594: This could be rewritten to:

```
  github_owner       = var.helm_exporter_github_owner != null ? var.helm_exporter_github_owner : var.platform_fluxcd_github_owner
  repo_name          = var.helm_exporter_repo_name != null ? var.helm_exporter_repo_name : var.platform_fluxcd_repo_name
  repo_branch        = var.helm_exporter_repo_branch != null ? var.helm_exporter_repo_branch : var.platform_fluxcd_repo_branch
```

This means that if the three variables are not defined, it will use the similar variables for Flux in general.

avnes:
For some reason I am not able to install the Helm-install in my sandbox. Could be a problem on my side. I will need to check again next week.

avnes:
When patching the helm chart values file (from terraform) we need to add values like this:

https://github.com/dfds/helm-exporter/blob/main/values.yaml

Where the list of charts includes all charst we got:

helm list -A

_sub/monitoring/helm-exporter/vars.tf
Ranixa: Thanks, added!

compute/k8s-services/main.tf
Ranixa: Okie dokie, you will have to explain to me why we wouldn't define these in the .hcl but i have added them.

rifisdfds (approved):

avnes (approved):
LGTM, and also seems to work fine in my sandbox.

PR #575, opened 2022-12-27
Created new IAM role for Capability accounts

SEQUOIIA:
The new role in each Capability account will be used for what is currently the 'aws2k8s' service. A similar role will need to be created for any non-Capability accounts that is being used for getting access to EKS.


security/org-account-context/dependencies.tf
```diff
@@ -13,6 +13,17 @@ data "aws_iam_policy_document" "assume_role_policy_self" {
   }
 }
 
+data "aws_iam_policy_document" "assume_role_policy_selfservice" {
+  statement {
+    actions = ["sts:AssumeRole"]
+
+    principals {
+      type        = "AWS"
+      identifiers = ["arn:aws:iam::${var.shared_account_id}:role/aad-aws-sync"]
```
emil2k, line 22: Where does this role get created?

SEQUOIIA: It doesn't. Not yet anyways. It would have to be done separately. Since this repo is public, I'm not going to disclose it here.

emil2k (approved):

rifisdfds (approved):

PR #593, opened 2023-01-05
removed deny cw logs read access on non eu

samidbb:
it's a challenge for people to not be able to view logs on global resources like cloudwatch lambda functions!


security/org-capability-root/dependencies.tf
```diff
@@ -19,8 +19,6 @@ locals {
                 "ce:Get*",
                 "ce:List*",
                 "cloudfront:*",
-                "cloudwatch:Get*",
-                "cloudwatch:List*",
```
emil2k, line 23: These are in a `NotAction` statement which means the `Deny` won't be applied to these actions, removing these will make the policy less permissive, but that doesn't seem to be what you are trying to do.

emil2k (changes requested):
I'm not sure these changes are going to accomplish the stated goal.

security/org-capability-root/dependencies.tf
rifisdfds: Unless i'm misunderstanding the intention, the addition should now allow the cloudwatch get and list actions in non-eu regions which I think is the intention so this change should be fine. @samidbb can you confirm?

emil2k: I read it as removing it from the `NotAction` block removes the exception, so now the `Deny` will be applied to the `cloudwatch:Get/List*` actions in non-eu regions.

rifisdfds: Sorry yes you are correct, clearly I wasn't awake this morning. I read this change as an addition rather than removal. The red and - should have given it away really

emil2k (changes requested):
I don't see any new changes since my last review here: https://github.com/dfds/infrastructure-modules/pull/593#pullrequestreview-1240541069

Am I missing something?

security/org-capability-root/dependencies.tf
samidbb: It should remove the cloudwatch from the blacklist, so it allows us to read cw data on non-eu region so I don't understand the confusion about it?

emil2k: I think it is the opposite of a blacklist, as mentioned above - let me know if you think that is wrong.

samidbb: @emil2k it's opposite of a blacklist because it removes the two actions from the deny list

PR #594, opened 2023-01-05
Bump QA's fluxcd version to v0.38.2

emil2k:


emil2k:
This has passed the QA twice now, it should be good to go.

PR #597, opened 2023-01-06
Create secret-detection.yml

jdsmithit:
Add in secret detection that will upload to the security tab.

Also bump the sarif upload version in tfsec


avnes (approved):
LGTM

PR #599, opened 2023-01-11
Upgrade QA to v1.23

aleksandra-korczak:


emil2k:
@aleksandra-korczak Note that you need to do a rollover to test the apply of this branch as the AMIs are switched, and the QA doesn't currently do that.

PR #652, opened 2023-01-27
Update Traefik to v20.8.0 in QA

emil2k:
Ability to update the traefik proxy with a blue/green deployment.


avnes (approved):

_sub/compute/k8s-traefik-flux/dependencies.tf
```diff
@@ -62,7 +62,7 @@ locals {
     "apiVersion" = "kustomize.config.k8s.io/v1beta1"
     "kind"       = "Kustomization"
     "resources" = [
-      "https://github.com/dfds/platform-apps/apps/${var.deploy_name}"
+      "https://github.com/dfds/platform-apps/apps/${var.deploy_name}?ref=add-traefik-variant" # TODO(emil): remove this after testing
```
emil2k, line 65: Remove this before merging.

compute/k8s-services/main.tf
```diff
@@ -150,37 +179,48 @@ module "traefik_alb_auth" {
...
+  # Green variant
+  deploy_green_variant            = var.traefik_alb_auth_deploy && var.traefik_green_variant_flux_deploy
+  green_variant_target_http_port  = var.traefik_green_variant_flux_http_nodeport
+  green_variant_target_admin_port = var.traefik_green_variant_flux_admin_nodeport
+  green_variant_health_check_path = "/ping"
+  green_variant_weight            = var.traefik_green_variant_flux_weight
 }
 
-module "traefik_alb_auth_dns" {
+module "traefik_alb_auth_dns_for_grafana" {
```
avnes, line 199: We must also keep the DNS alias for `["internal.${var.eks_cluster_name}"]`, because it is used as a catch-all rule for anything you want to test with AUTH, and not having to create a new DNS alias for it.

compute/k8s-services/vars.tf
```diff
@@ -704,6 +704,15 @@ variable "crossplane_confluent_clusters_endpoints" {
...
+# version without downtime.
+
+# TODO(emil): Rename the original Traefik instance resources and variables to
+# specify that they refer to the "blue" variant after the "blue" instance is
+# destroyed.  This is to avoid downtime or having to reimport resources due to
+# renaming.
+
 variable "traefik_flux_github_owner" {
   type        = string
   description = "Name of the Treaefik Flux repo Github owner (previously: organization)"
```
avnes, line 718: Typo. Not actually introduced by you, but please fix since it is the same file as you are already chaning.

'Traefik' not 'Treaefik'

avnes (changes requested):

compute/k8s-services/vars.tf
emil2k: Fixed.

avnes (approved):

PR #678, opened 2023-02-15
Fluentd with containerd

emil2k:
- Configure fluentd based on runtime
- Concatenates partial logs emitted by containerd using the fluent-plugin-concat plugin
- Validation for container runtime values
- If either of the node groups in the EKS cluster has a `dockerd` container runtime it will configure fluentd to collect from `dockerd` otherwise it will be configured to collect from `containerd`


emil2k:
Tested in my sandbox:
- Collection with dockerd and containerd as the container runtime.
- Concatenation of large log lines (>16 kB) with containerd as runtime.

avnes (approved):
LGTM, and worked fine in my sandbox (that is running with containerd)

PR #691, opened 2023-02-21
Use pessimistic version locking. Also fix some tfsec issues

avnes:
- Use pessimistic verion locking. Also some tfsec fixes
- Re-enable tfsec

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,8 +3,8 @@ repos:
   rev: v1.59.0
   hooks:
     - id: terraform_fmt
-      #    - id: terraform_tfsec
-      #files: (\.tf|\.tfvars)$
+    - id: terraform_tfsec
+      files: (\.tf|\.tfvars)$
 - repo: git@github.com:Yelp/detect-secrets
   rev: v1.1.0
   hooks:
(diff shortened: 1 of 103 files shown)


emil2k (changes requested):
It's failing QA at the moment with:

```
╷
│ Error: Failed to query available provider packages
│ 
│ Could not retrieve the list of available versions for provider
│ hashicorp/aws: no available releases match the given constraints ~> 4.52.0,
│ ~> 4.54.0
╵
```

avnes:
> It's failing QA at the moment with:
> 
> ```
> ╷
> │ Error: Failed to query available provider packages
> │ 
> │ Could not retrieve the list of available versions for provider
> │ hashicorp/aws: no available releases match the given constraints ~> 4.52.0,
> │ ~> 4.54.0
> ╵
> ```

Isn't that sort of expected, since we on master allow "version = ">= 4.52.0"? 

Since master branch is applied before feature branch, this will fail until we merge.

emil2k:
I don't think this has to do with the master branch at all, I think it is because we use both the constraint `~> 4.54.0` and `~> 4.52.0` simultaneously within the repo and it cannot fulfill that.

We probably need to update the 52s to 54s.

emil2k (approved):

PR #709, opened 2023-03-01
Set client.authentication.k8s.io/v1beta1 as the default API version for authentication

avnes:


emil2k (approved):

emil2k:
Just wait for the QA run to go green.

PR #723, opened 2023-03-06
Don't use auth_alb with Grafana. Integrate directly with MS OAuth service

avnes:
- Use Grafana with OAuth integration
- Make sure kube-prometheus-stack is not depending on auth_alb


compute/k8s-services/vars.tf
```diff
@@ -983,12 +983,6 @@ variable "velero_flux_log_level" {
   }
 }
 
-variable "velero_flux_github_owner" {
```
SEQUOIIA, line 986: Seems out of scope for this PR, but I'll allow it 😂

SEQUOIIA (approved):
Added a comment for compute/k8s-services/vars.tf

PR #748, opened 2023-03-15
Add csi-gp3 as storage class and make it the default

avnes:


emil2k (approved):

emil2k:
It might be good to switch the QA configs to use this class (atlantis, prometheus, grafana): https://github.com/dfds/infrastructure-modules/blob/master/test/integration/eu-west-1/k8s-qa/services/terragrunt.hcl

PR #755, opened 2023-03-16
Use platform-apps with a branch

avnes:


_sub/compute/k8s-traefik-flux/dependencies.tf
```diff
@@ -62,7 +62,7 @@ locals {
     "apiVersion" = "kustomize.config.k8s.io/v1beta1"
     "kind"       = "Kustomization"
     "resources" = [
-      "https://github.com/dfds/platform-apps/apps/${var.deploy_name}"
+      "${var.gitops_apps_repo_url}/apps/${var.deploy_name}?ref=${var.gitops_apps_repo_branch}"
```
emil2k, line 65: Shouldn't you run this through `urlencode`?

emil2k: Does a branch with a forward slash work?

_sub/compute/k8s-fluxcd/vars.tf
```diff
@@ -23,6 +23,18 @@ variable "github_owner" {
   description = ""
 }
 
+variable "gitops_apps_repo_url" {
+  type        = string
+  default     = ""
```
emil2k, line 28: Would it not be better if these didn't have a default value? This default does not seem valid.

emil2k (changes requested):
Just some questions.

_sub/compute/k8s-traefik-flux/dependencies.tf
avnes: Yes, it works, it has been tested in my sandbox. Here is an output from my sandbox:

`platform-apps-sources                     	feature/pss-psa@sha1:9d3ee061	False    	True 	Applied revision: feature/pss-psa@sha1:9d3ee061`

_sub/compute/k8s-fluxcd/vars.tf
avnes: These two vars have a default, but that is in the main module. The logic was that the gitops_apps_repo_url and gitops_apps_repo_branch could potentially be used by other main modules or by non-DFDS users.

emil2k (approved):

PR #759, opened 2023-03-17
Only apply target group block on multiple variants

Mik-Nord:


_sub/compute/eks-alb/main.tf
emil2k, line 1: There is two of these blocks, another one in the `anon` alb.

_sub/compute/eks-alb/main.tf
```diff
@@ -88,31 +88,34 @@ resource "aws_lb_listener" "traefik" {
...
-            {
-              arn    = aws_lb_target_group.traefik_blue_variant[0].arn
-              weight = var.blue_variant_weight
-            }
-          ] : [],
-          var.deploy_green_variant ? [
-            {
-              arn    = aws_lb_target_group.traefik_green_variant[0].arn
-              weight = var.green_variant_weight
+    target_group_arn = var.deploy_blue_variant && var.deploy_green_variant ? null : (var.deploy_blue_variant ? var.deploy_blue_variant : var.deploy_green_variant)
```
emil2k, line 91:
```suggestion
    target_group_arn = var.deploy_blue_variant && var.deploy_green_variant ? null : try(aws_lb_target_group.traefik_blue_variant[0].arn, aws_lb_target_group.traefik_green_variant[0].arn)
```

emil2k (changes requested):

PR #781, opened 2023-03-28
Tune atlantis resource requests/limits.

emil2k:
Current resource requests are excessive, especially on a sandbox.

### DIFF
--- a/_sub/compute/helm-atlantis/values/values.yaml
+++ b/_sub/compute/helm-atlantis/values/values.yaml
@@ -4,9 +4,11 @@ image:
   pullPolicy: IfNotPresent
 
 resources:
+  requests:
+    memory: 100Mi
+    cpu: 50m
   limits:
-    memory: 2Gi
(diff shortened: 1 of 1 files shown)


_sub/compute/helm-atlantis/values/values.yaml
```diff
@@ -4,9 +4,11 @@ image:
   pullPolicy: IfNotPresent
 
 resources:
+  requests:
+    memory: 100Mi
+    cpu: 50m
   limits:
-    memory: 2Gi
-    cpu: 1
+    memory: 100Mi
```
samidbb, line 11: why there is no limit on the CPU usage?

emil2k: I was following some advice that specifying CPU limits is actually a bad idea in Kubernetes: https://home.robusta.dev/blog/stop-using-cpu-limits

samidbb (approved):

PR #782, opened 2023-03-28
remove adfs from org-account-context

rifisdfds:
This will remove the ADFS items from org-account-context (used by aws-account-manifests)


SEQUOIIA (approved):
batman

PR #783, opened 2023-03-29
Add support for the new service control policies without destroying the

avnes:
old ones yet.


security/org-policies/dependencies.tf
```diff
@@ -0,0 +1,341 @@
...
+        "logs:DeleteLogGroup",
+        "logs:DeleteLogStream"
+      ],
+      "Resource": "*"
+    },
+    {
+      "Sid": "DenyDisablingCloudTrail",
+      "Action": [
+        "cloudtrail:StopLogging",
+        "cloudtrail:DeleteTrail"
```
emil2k, line 141: Maybe `UpdateTrail` should also be disallowed as someone could modify the events that are logged to effectively disable the trail.

security/org-policies/dependencies.tf
```diff
@@ -0,0 +1,341 @@
...
+      "Action": [
+        "access-analyzer:DeleteAnalyzer"
+      ],
+      "Resource": "*",
+      "Effect": "Deny"
+    },
+    {
+      "Sid": "DenyDisablingEditingAWSConfig",
+      "Action": [
+        "config:DeleteConfigRule",
```
emil2k, line 157: Maybe `PutConfigRule` also to prevent updates.

security/org-policies/dependencies.tf
```diff
@@ -333,7 +333,51 @@ POLICY
...
+                    "aws:PrincipalArn": [
+                        "arn:aws:iam::*:role/OrgRole",
+                        "arn:aws:iam::*:role/aws-reserved/sso.amazonaws.com/*/AWSReservedSSO_CloudAdmin_*",
+                        "arn:aws:iam::*:role/aws-reserved/sso.amazonaws.com/*/AWSReservedSSO_Billing_*"
+                    ]
+                }
+            }
+        },
+        {
+            "Sid": "DenyBillingLegacyRemoveAfterJuly2023",
```
emil2k, line 358: Can you put a note here on what the plan is with this policy? Why is it being removed?

emil2k (changes requested):
Looks good for the most part, just minor comments/suggestions. The blocking change here is that some changes that @samidbb recently put in recently https://github.com/dfds/infrastructure-modules/pull/593 might be rolled back by this.

Also looking at the design doc there is a section about blocking services competing with current offerings (i.e. code start, code commit, etc), but I did not see anything in this PR to deny those and don't see (quick glance) any issues in the milestone.

_sub/security/org-service-control-policy/main.tf
```diff
@@ -0,0 +1,17 @@
+# TODO: This is V2 of the org-policy sub module that needs to be kept a bit more for legacy reasons
```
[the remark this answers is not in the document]
avnes: Regarding #593 I will check that with @jdsmithit . I am not sure if he has approved Sami's change or not.

Regarding the TODO, I have an assigned ticket for that, which will be done after Easter holidays:
https://github.com/dfds/cloudplatform/issues/1502

security/org-policies/dependencies.tf
avnes: Jan want to deny all teams to view their billing data. I think the idea is to force them to use a FinOps tool we will subscribe to for 1 year.

avnes: Also, the IAM policies for billings are being replaced with new ones. Ref: https://docs.aws.amazon.com/cost-management/latest/userguide/migrate-granularaccess-whatis.html#migrate-user-permissions

emil2k: I see, I think it would be good if these types of plans were documented near the relevant code as comments.

security/org-policies/dependencies.tf
```diff
@@ -0,0 +1,341 @@
...
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Sid": "DenyDeletingVPCFlowLogs",
+      "Effect": "Deny",
+      "Action": [
+        "ec2:DeleteFlowLogs",
+        "logs:DeleteLogGroup",
+        "logs:DeleteLogStream"
```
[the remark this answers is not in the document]
avnes: @jdsmithit will consider, and get back to us.

security/org-policies/dependencies.tf
avnes: Thanks, it has been added now.

security/org-policies/dependencies.tf
```diff
@@ -0,0 +1,341 @@
...
+      "Effect": "Deny",
+      "Action": [
+        "iam:*"
+      ],
+      "Resource": [
+        "arn:aws:iam::*:role/OrgRole"
+      ]
+    },
+    {
+      "Sid": "DenyAllOutsideEU",
```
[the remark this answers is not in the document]
avnes: Thanks, it has been added now.

security/org-policies/dependencies.tf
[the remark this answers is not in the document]
avnes: Regarding CodeStart etc, these are future policy addiitions that have not been carved out yet. We will start work on those after Easter holidays.

emil2k (approved):

security/org-policies/dependencies.tf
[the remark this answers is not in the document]
avnes: After speaking with @jdsmithit we have opted to deny people to delete CloudWatch logs - unless it is done by the EKSAdmin role. I also added the OrgRole after speaking with John, so we as admins also can switch to accounts and delete cloudwatch logs. 

Hence the DenyDeletingVPCFlowLogs policy has been reduced to only handle VPC Flow Logs, and a new policy DenyDeletingCloudWatchLogs has been added.

PR #784, opened 2023-03-29
remove adfs from org-account-assume

rifisdfds:
This PR will remove the ADFS provider and dependent roles from the org-account-assume module


security/org-account-assume/main.tf
```diff
@@ -101,56 +91,3 @@ resource "aws_iam_role_policy" "prime-admin" {
   provider = aws.workload
 }
 
-# Create cloud-engineer role for the workload account
-resource "aws_iam_role" "cloudengineer_role" {
-  name                 = var.cloudengineer_iam_role_name
```
emil2k, line 106: Is this var used anywhere else?

security/org-account-assume/main.tf
```diff
@@ -101,56 +91,3 @@ resource "aws_iam_role_policy" "prime-admin" {
...
-  max_session_duration = 28800
-
-  provider = aws.workload
-}
-
-# Policy inline to cloud-engineer role
-resource "aws_iam_role_policy" "cloudengineer" {
-  name   = "CloudEngineer"
-  role   = aws_iam_role.cloudengineer_role.id
-  policy = module.iam_policies.cloudengineer
```
emil2k, line 118: Is this policy used anywhere else? Maybe it can be cleaned up also.

security/org-account-assume/main.tf
```diff
@@ -101,56 +91,3 @@ resource "aws_iam_role_policy" "prime-admin" {
...
-  role       = aws_iam_role.cloudengineer_role.name
-  policy_arn = "arn:aws:iam::aws:policy/job-function/SupportUser"
-
-  provider = aws.workload
-}
-
-
-# Create cloud-admin role for the workload account
-resource "aws_iam_role" "cloudadmin_role" {
-  name                 = var.cloudadmin_iam_role_name
```
emil2k, line 141: This one too.

emil2k (approved):
Looks good, but there might be a couple of things you can clean up before merging it in.

security/org-account-assume/main.tf
rifisdfds: Nope, spotted and tested removal of these before lunch but hadn't pushed. Good that you've got a keen eye!

security/org-account-assume/main.tf
rifisdfds: Yeah they are, can't remove it until other PR's are through

PR #791, opened 2023-03-31
Automated inactivity triggered EKS/ELB clean up for sandboxes.

emil2k:
Creates a CloudWatch alarm that aims to detect inactivity. The alarm will trigger the deletion of the EKS cluster and ELB load balancers. It will fire after a workday (excludes weekends) where sandbox ELB target groups have 0 targets. CloudWatch limits the evaluation period to 1 day so this was the best workaround to avoid triggering the inactivity clean-up due to normal activity during weekends.

To reboot the resources, one needs to first run `terragrunt apply -target=module.eks_cluster` to recreate the EKS control plane. Afterwards, run the normal `terragrunt apply`s to recreate the rest of the resources along with the ELBs.

To opt out of this clean up, use the variables
`disable_inactivity_cleanup`. Use this sparingly, not as a default.

### DIFF
--- a/_sub/compute/eks-alb-auth/outputs.tf
+++ b/_sub/compute/eks-alb-auth/outputs.tf
@@ -2,6 +2,14 @@ output "alb_fqdn" {
   value = element(concat(aws_lb.traefik_auth[*].dns_name, [""]), 0)
 }
 
+output "alb_name" {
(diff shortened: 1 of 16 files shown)


samidbb (approved):
lgtm

PR #810, opened 2023-04-12
Enable postgres storage autoscaling

jakob-nielsen-dfds:
Enable Postgres storage autoscaling by defining the optional `max_allocated_storage` variable, as per: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/db_instance#max_allocated_storage

By default, it is set to 0, which disables autoscaling.


emil2k (approved):
Looks good to me. Thank you for the contribution.

PR #814, opened 2023-04-14
postgres.output.connection_string: flag as sensitive

izotomas:
if the password is sensitive, `connection_string` should be protected as well - to prevent password leak


emil2k (approved):

emil2k:
Thank you for the contribution.

PR #825, opened 2023-04-24
1522 log alarm

Mik-Nord:


_sub/monitoring/cloudwatch-alarms/log-anomaly/versions.tf
```diff
@@ -0,0 +1,9 @@
+terraform {
+  required_version = ">= 1.3.0, < 2.0.0"
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = "~> 4.63.0"
```
avnes, line 6: This needs to be 4.64.0

avnes (changes requested):

_sub/monitoring/cloudwatch-alarms/log-anomaly/main.tf
```diff
@@ -0,0 +1,39 @@
...
+      dimensions = {
+        LogGroupName = "/k8s/hellman/kube-system"
+      }
+    }
+  }
+
+  metric_query {
+    id          = "ad1"
+    label       = "IncomingLogEvents (expected)"
+    expression  = "ANOMALY_DETECTION_BAND(m1, 1)"
```
emil2k, line 28: Is 1 standard deviation enough to trigger an alarm? Might be too sensitive.

emil2k (approved):

emil2k (changes requested):
Need to update the AWS provider as Audun said. I meant to just post a comment but accidentally approved the PR initially.

_sub/monitoring/cloudwatch-alarms/log-anomaly/main.tf
Mik-Nord: According to the graph, we never don't recieve logs unless there's a problem. But it's hard to find a sweet spot. Also, since we can't bind it to x amount of nodes, the deviation actually means more than 1 node are not sending log events.

_sub/monitoring/cloudwatch-alarms/log-anomaly/versions.tf
Mik-Nord: Applied change request

_sub/monitoring/cloudwatch-alarms/log-anomaly/main.tf
emil2k: The standard deviation set at 1 essentially means that 1 out of 6 times the threshold might be crossed due to random variation. At 2, it should be down to 1 out of 40 times at 3 it should be at about 1 out of 200.

Since you are requiring two consecutive samples to fire the alarm the probability of the alarm firing due to random variation should be 1/6 * 1/6 = 1 out of 36. If the evaluation period is 5 minutes there are 288 periods within a day, so you would expect this to fire due to randomness ~8 a day.

If I am understanding correctly, it might be too sensitive, but give it a try I guess ...

conspiracy

emil2k (approved):

_sub/monitoring/cloudwatch-alarms/log-anomaly/main.tf
Mik-Nord: The thing is that it measures the amount of logs and creates a buffer around that, so 1 node going down would not result in 1 deviation. Look here:
image

Mik-Nord: Okay looking further into it, it might be too sensitive.
image
This would have fired twice then

PR #829, opened 2023-04-25
output id from ou in org-capability-root module

rifisdfds:
This PR will allow the output of OU ID from the org-capability-root module. This allows the use of the ID from terragrunt/terraform calling the org-capability-root module so we can refer to OU's by reference rather than hard-coding strings

### DIFF
--- a/security/org-capability-root/outputs.tf
+++ b/security/org-capability-root/outputs.tf
@@ -1,0 +1,3 @@
+output "id" {
+  value = module.capability_ou.id
+}


SEQUOIIA (approved):
https://user-images.githubusercontent.com/1633308/234286368-48855dca-9eb2-4955-a442-5a7f878ad47c.mp4

jdsmithit (commented):
approve

PR #830, opened 2023-04-25
Atlantis refactor and prep for staging

emil2k:
- Move to [repo level `atlantis.yaml` config](https://www.runatlantis.io/docs/repo-level-atlantis-yaml.html)
- Make the `helm-atlantis` module more reusable by not blindly passing through variables from services
- Enable staging environment with new variables passed into services
- Removed unused variables:  `atlantis_platform_fluxcd_github_token`, `atlantis_access_key_master`, `atlantis_secret_key_master`
- Split storage of Atlantis credentials in SSM into two parts to avoid `jsonencode` (one then has to decode to use the credential)
- Switch the default storage class for Atlantis from `csi-gp2` to `csi-gp3`.
- Move away from using deprecated `orgWhitelist` in the Atlantis Helm chart

--- 

Note that one will need to remap the environment variables like so in the `atlantis.yaml` in different workflows depending on the usage, like so:

```yaml
        - env:
            name: AWS_ACCESS_KEY_ID
            command: "echo $STAGING_AWS_ACCESS_KEY_ID"
```

### DIFF
--- a/compute/k8s-services/main.tf
+++ b/compute/k8s-services/main.tf
@@ -537,6 +537,7 @@ module "atlantis" {
     arm_client_id                                           = var.atlantis_arm_client_id
     arm_client_secret                                       = var.atlantis_arm_client_secret
     tf_var_monitoring_kube_prometheus_stack_azure_tenant_id = var.monitoring_kube_prometheus_stack_azure_tenant_id
+    tf_var_platform_fluxcd_github_token                     = var.atlantis_platform_fluxcd_github_token
     tf_var_atlantis_github_token                            = var.atlantis_github_token
     tf_var_atlantis_platform_fluxcd_github_token            = var.atlantis_platform_fluxcd_github_token
   }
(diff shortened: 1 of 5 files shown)


compute/k8s-services/main.tf
```diff
@@ -502,37 +502,43 @@ module "platform_fluxcd" {
...
+  atlantis_image_tag  = var.atlantis_image_tag
+  atlantis_ingress    = var.atlantis_ingress
+  storage_class       = var.atlantis_storage_class
+  github_username     = var.atlantis_github_username
+  github_token        = var.atlantis_github_token
+  github_repositories = var.atlantis_github_repositories
+  webhook_url         = var.atlantis_ingress
+  webhook_events      = var.atlantis_webhook_events
+
+  environment_variables = {
```
avnes, line 521: nice

compute/k8s-services/vars.tf
```diff
@@ -525,6 +530,36 @@ variable "atlantis_image_tag" {
   default     = "latest"
 }
 
+variable "atlantis_storage_class" {
+  type        = string
+  description = "Storage class to use for persistent volume"
+  default     = "csi-gp2"
```
avnes, line 536: How about using csi-gp3 instead?

avnes (changes requested):

avnes (approved):

PR #839, opened 2023-04-28
Abillity to specify velero cron schedule.

emil2k:
Since sandbox/QA/staging environments scale down at 18 UTC, if we want to test Velero we need to be able to override the schedule time so that it happens before the node scale down.

### DIFF
--- a/compute/k8s-services/vars.tf
+++ b/compute/k8s-services/vars.tf
@@ -932,7 +932,7 @@ variable "velero_bucket_name" {
 
 variable "velero_cron_schedule" {
   type        = string
-  default     = null
+  default     = "0 0 * * *"
   description = "Cron-formatted scheduled time for the Velero backup."
 }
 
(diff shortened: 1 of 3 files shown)


compute/k8s-services/vars.tf
```diff
@@ -930,6 +930,12 @@ variable "velero_bucket_name" {
   description = "The name of the S3 bucket that contains the Velero backup"
 }
 
+variable "velero_cron_schedule" {
```
avnes, line 933: Did you forget to push? I don't see velero_cron_schedule being used anywhere

avnes (changes requested):

compute/k8s-services/vars.tf
emil2k: https://github.com/dfds/infrastructure-modules/pull/839/files#diff-9dcb2b34c331c83c7b15b8785fc38080f22e765d3119f854acc7848e7f669485R764

avnes: My eyes, my eyes :)

avnes (approved):

PR #841, opened 2023-05-01
use helm exporter 1.2.5 in QA

avnes:


emil2k (approved):

emil2k:
For stuff that affects only QA we can use the `release:norelease` label, as it should not produce a release.

PR #852, opened 2023-05-03
Restrict grafana dashboard sidecar namespace selector to monitoring

SEQUOIIA:


silviucalin (approved):

emil2k (approved):
Isn't the sidecar running in the monitoring namespace already? 

>     # Otherwise the namespace in which the sidecar is running will be used.
https://github.com/grafana/helm-charts/blob/main/charts/grafana/values.yaml#L928

I guess this is worth a try still.

SEQUOIIA:
> Isn't the sidecar running in the monitoring namespace already?
> 
> > ```
> > # Otherwise the namespace in which the sidecar is running will be used.
> > ```
> 
> https://github.com/grafana/helm-charts/blob/main/charts/grafana/values.yaml#L928
> 
> I guess this is worth a try still.

It's not where the sidecar is deployed, but rather where it looks for configmaps.

emil2k:
Yes, I just mean it says that if that setting is unset (as it was) it should default the search to the namespace where it is running.

SEQUOIIA:
Ah, yeah. That is indeed what is written, but the behaviour seen when deployed is different 😬

PR #856, opened 2023-05-04
Updating QA control plane to 1.26

silviucalin:
This is normal procedure in preparation for production upgrade.


emil2k (approved):
Looks fine. Note that this change is not reversible so QA will fail on all other branches that are still on 1.25, so get this in as soon as possible or nuke QA.

PR #866, opened 2023-05-08
SCP exception for OrgRole to modify/delete CloudTrail resources.

emil2k:


jdsmithit:
For context what will be doing the action for these exceptions is it a lambda or just cloudwatch/Event bridge?

emil2k:
@jdsmithit It will be the accounts pipeline.

jdsmithit (approved):

PR #869, opened 2023-05-09
Setup a Github Actions TFLint

silviucalin:
This is in relation with the issue on the board raised by Audun.

### DIFF
--- a/.github/workflows/tflint.yml
+++ b/.github/workflows/tflint.yml
@@ -1,0 +1,42 @@
+name: Lint
+
+on:
+  push:
+    branches: [ "master", "main" ]
+  pull_request:
+    branches: [ "master", "main" ]
+
+jobs:
+  tflint:
+    runs-on: ${{ matrix.os }}
+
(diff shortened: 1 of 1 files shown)


.github/workflows/tflint.yml
```diff
@@ -0,0 +1,42 @@
...
+  pull_request:
+    branches: [ "master", "main" ]
+
+jobs:
+  tflint:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
emil2k, line 15: Is this necessary? What is OS-specific about running tflint?

emil2k (approved):

.github/workflows/tflint.yml
silviucalin: I'd say it isn't, maybe just use ubuntu-latest rather.
I guess it's to fit any type of GitHub runner available.

PR #887, opened 2023-05-16
Add support for Certero

avnes:


security/org-account-context/dependencies.tf
```diff
@@ -35,3 +35,14 @@ data "aws_iam_policy_document" "shared_role_cap_acc" {
...
   }
 }
+
+data "aws_iam_policy_document" "assume_role_policy_master_account" {
+  statement {
+    actions = ["sts:AssumeRole"]
+
+    principals {
+      type        = "AWS"
+      identifiers = ["arn:aws:iam::${var.master_account_id}:root"]
```
emil2k, line 45: We should try to limit these trust relationships.

emil2k (approved):

security/org-account-context/dependencies.tf
avnes: Well, it is a requirement that the vendor Certero has from connecting from our master account to each organization account.

emil2k: Yeah, but we can limit it to the user/role they will utilize instead of `root`.

_sub/security/iam-policies/main.tf
```diff
@@ -213,3 +213,26 @@ data "aws_iam_policy_document" "trusted_account" {
...
+    resources = ["*"]
+    actions   = local.certero_actions_for_org_accounts
+  }
+}
+
+data "aws_iam_policy_document" "certero_endpoint_for_master_account" {
+  statement {
+    sid       = "CerteroEndpoint"
+    effect    = "Allow"
+    resources = ["*"]
```
emil2k, line 235: It might also be good to limit the `AssumeRole` action to only the `CerteroRole` in the org accounts.

emil2k (approved):
It looks good to me, just a minor suggestion to improve the least privilige.

PR #893, opened 2023-05-17
Add support for S3 bucket for kubeconfig being private

avnes:
https://github.com/dfds/cloudplatform/issues/940

### DIFF
--- a/storage/s3-eks-public/vars.tf
+++ b/storage/s3-eks-public/vars.tf
@@ -31,3 +31,19 @@ variable "additional_tags" {
   type        = map(string)
   default     = {}
 }
+
+variable "acl" {
+  description = "The ACL is the S3 bucket security settings"
+  type        = string
+  default     = "public-read"
+
+  validation {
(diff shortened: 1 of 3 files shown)


emil2k (approved):

storage/s3-eks-public/vars.tf
```diff
@@ -31,3 +31,19 @@ variable "additional_tags" {
...
+  type        = string
+  default     = "public-read"
+
+  validation {
+    condition     = (var.acl == "public-read" || var.acl == "private")
+    error_message = "Bucket ACL must be either public-read private. Defaults to public-read. Should be private for sandboxes."
+  }
+}
+
+variable "eks_is_sandbox" {
```
emil2k, line 46: We should probably use this in the `sandbox-template`.

avnes: Yes, we must use this in the template, because it needs to be defined at the highest level, in account.tfvars :)

avnes: Done

PR #895, opened 2023-05-18
feat: add datadog aws integration

rifisdfds:
This PR adds the datadog aws integration submodule
The org-account, org-account-assume and org-account-context main modules are configured to use the module in a backwards-compatible manner, disabled by default.


emil2k (approved):

samidbb (approved):
tested on sandbox

PR #904, opened 2023-05-22
Harden default security groups

emil2k:
Disables egress/ingress in the default security groups for hardened accounts's VPC and the EKS cluster VPC.


avnes (approved):

aleksandra-korczak (approved):
LGTM

PR #908, opened 2023-05-23
Only allow non-global services from eu-central-1 and eu-west-1

avnes:
https://github.com/dfds/cloudplatform/issues/1527 and https://dfdsit.statuspage.io/incidents/hxg95tvphjfl


silviucalin (approved):
Let's go!

emil2k (approved):

PR #909, opened 2023-05-23
Determine admin user to manage CMK.

emil2k:
Requires use of aws_iam_session_context to retrieve the role or user based ARN.

Need to make a change to the SCP to allow iam:GetRole on the OrgRole to utilize that resource.


security/org-policies/policies/RestrictivePolicy.json
```diff
@@ -106,12 +106,8 @@
         {
             "Sid": "DenyOrgRoleModification",
             "Effect": "Deny",
-            "Action": [
-                "iam:*"
-            ],
-            "Resource": [
-                "arn:aws:iam::*:role/OrgRole"
-            ]
+            "NotAction": ["iam:GetRole"],
```
emil2k, line 109: @avnes This might block more than the original `iam:*` actions (`sts:*`), so I'm not really sure about using this `NotAction`.

Could we come up with a list of actions that we want to explicitly deny here instead of the original `iam:*` since the intent is to only prevent modification of the role?

avnes (changes requested):

avnes (approved):

PR #924, opened 2023-05-30
Add support for Atlantis to pre-prime pipeline

avnes:


emil2k (approved):
Should we maybe bump up the resources for Atlantis since it is servicing more pipelines? https://github.com/dfds/infrastructure-modules/blob/master/_sub/compute/helm-atlantis/values/values.yaml#L6-L12

avnes:
> Should we maybe bump up the resources for Atlantis since it is servicing more pipelines? https://github.com/dfds/infrastructure-modules/blob/master/_sub/compute/helm-atlantis/values/values.yaml#L6-L12

I'll keep an eye on it.

PR #926, opened 2023-05-30
Add support for tuning Atlantis

avnes:


_sub/compute/helm-atlantis/vars.tf
```diff
@@ -57,6 +57,24 @@ variable "data_storage" {
...
+
+variable "resources_requests_memory" {
+  type        = string
+  default     = "1536Mi"
+  description = "Memory resources requests size"
+}
+
+variable "resources_limits_memory" {
+  type        = string
+  default     = null
```
emil2k, line 74:
```suggestion
  default     = "1536Mi"
```

emil2k (changes requested):

compute/k8s-services/vars.tf
```diff
@@ -619,6 +619,23 @@ variable "atlantis_staging_aws_secret" {
...
+
+variable "atlantis_resources_requests_memory" {
+  type        = string
+  default     = "1536Mi"
+  description = "Memory resources requests size"
+}
+
+variable "atlantis_resources_limits_memory" {
+  type        = string
+  default     = null
```
emil2k, line 636:
```suggestion
  default     = "1536Mi"
```

emil2k (commented):
Looks good just need to adjust memory limits.

compute/k8s-services/vars.tf
avnes: This is on purpose. There is a local variable in the sub module that checks if resources_limits_memory is different than null or not. If it is null, then it will use the same value as for resources_requests_memory

test/integration/eu-west-1/k8s-qa/services/terragrunt.hcl
```diff
@@ -189,6 +189,9 @@ inputs = {
   atlantis_image_tag     = "0.0.38"
   atlantis_data_storage  = "1Gi"
 
+  atlantis_resources_requests_cpu    = "10m"
+  atlantis_resources_requests_memory = "512Mi"
```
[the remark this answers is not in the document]
avnes: No, in this case that it is missing, the sub module will take care of it, and set atlantis_resources_limits_memory to the same value as atlantis_resources_requests_memory. That is part of the test

_sub/compute/helm-atlantis/vars.tf
emil2k: Oh ok, missed that part.

emil2k (approved):

PR #931, opened 2023-05-31
Exception for OrgRole to be able to delete flow logs.

emil2k:
Sorry for the diff, but my JSON formatter wants it to look like this and I was not able to find a way to modify it.


samidbb (approved):
more formatting noise :)

PR #947, opened 2023-06-06
Add DenyResourceExplorer, but allow it for OrgRole and AWS Account Manifests pipeline - in any region

avnes:


samidbb (approved):
great! thanks

PR #948, opened 2023-06-06
Remove datadog from main modules

avnes:
- Remove datadog from org-account
- Remove datadog from org-account-assume
- Remove datadog from org-account-context
- Remove datadog from k8s-services


SEQUOIIA:
I can't quite recall. If you remove the provider before removing provider-specific resources from the state, can it actually remove those resources?

SEQUOIIA (approved):

PR #956, opened 2023-06-09
Limit denying resource explorer policies to actions which create/modify resources.

emil2k:
…fy resources.


avnes (approved):

avnes (changes requested):
LGtM, but Git conflict

avnes (approved):

PR #976, opened 2023-06-15
Set security contact for hardened accounts.

emil2k:
Depends on #975


security/org-account-context/main.tf
```diff
@@ -518,6 +518,17 @@ module "config_local_2" {
   }
 }
 
+resource "aws_account_alternate_contact" "security" {
+  count                  = var.harden ? 1 : 0
+  alternate_contact_type = "SECURITY"
+  name                   = "Cloud Engineering"
+  title                  = "Cloud Engineering"
+  email_address          = "cloud.engineering+${var.name}@dfds.com"
```
avnes, line 526: This is a public repo, and in order to avoid unnecessary spam, I think this should be a variable with no default values.

security/org-account-context/main.tf
```diff
@@ -518,6 +518,17 @@ module "config_local_2" {
...
+  phone_number           = "+4533423456"
```
avnes, line 527: Phone number must also be a variable without a default value.

avnes (changes requested):

security/org-account-context/main.tf
emil2k: Make sense, addressed.

security/org-account-context/main.tf
emil2k: Addressed.

avnes (approved):

PR #996, opened 2023-06-22
Add github oidc provider

aleksandra-korczak:


silviucalin (approved):

_sub/security/iam-github-oidc-provider/main.tf
```diff
@@ -0,0 +1,48 @@
+resource "aws_iam_openid_connect_provider" "github" {
+  client_id_list  = ["sts.amazonaws.com"]
+  thumbprint_list = ["6938fd4d98bab03faadb97b34396831e3780aea1"]
```
avnes, line 3: Would it makes sense to supply the URL and thumbprint as variables?
I guess at least the thumbprint will change over time (when a new cert is issued)

avnes (approved):
LGTM, but one comment

_sub/security/iam-github-oidc-provider/main.tf
aleksandra-korczak: Yeah I thought about it for a while but I don't think these are going to change very often:
https://github.blog/changelog/2022-01-13-github-actions-update-on-oidc-based-deployments-to-aws/

PR #1006, opened 2023-06-23
Add support for latest goldpinger. 

avnes:
A new default version needed to be set in compute/services/vars.tf or else deployment would fail on systems where goldpinger has an old version installed from a previous helm repository

### DIFF
--- a/_sub/compute/helm-goldpinger/main.tf
+++ b/_sub/compute/helm-goldpinger/main.tf
@@ -1,7 +1,7 @@
 resource "helm_release" "goldpinger" {
   name          = "goldpinger"
   chart         = "goldpinger"
-  repository    = "https://charts.helm.sh/stable"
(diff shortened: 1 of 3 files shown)


_sub/compute/helm-goldpinger/main.tf
```diff
@@ -1,7 +1,7 @@
 resource "helm_release" "goldpinger" {
   name          = "goldpinger"
   chart         = "goldpinger"
-  repository    = "https://charts.helm.sh/stable"
+  repository    = "https://okgolove.github.io/helm-charts/"
```
emil2k, line 4: Make sure to update the Helm exporter vars when this is released.

emil2k (approved):

_sub/compute/helm-goldpinger/main.tf
avnes: Thanks

PR #1019, opened 2023-06-28
Remove-remaining-unused-noneu-providers

samidbb:


avnes (approved):
LGTM, but why comment out the resources instead of deleting them?

emil2k (approved):

samidbb:
> LGTM, but why comment out the resources instead of deleting them?

Because I want to see which providers I will have to remove in the the PR that will follow this one

PR #1064, opened 2023-07-24
sidecar updates for grafana alerts from CM

samidbb:
Enable using side car to load alert rules like we do with dashboards


samidbb:
replaced by this PR https://github.com/dfds/infrastructure-modules/pull/1065

PR #1082, opened 2023-08-11
Added norelease label automation and refactored renovate configurations

Mik-Nord:
Renovatebot do not have sufficient AND/OR operators when it comes to matching file paths. So a different GH action was required. When we apply our own custom labels, Renovatebot cannot automerge unfortunately.

[Ops ticket](https://github.com/dfds/cloudplatform/issues/1865)

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -1,0 +1,63 @@
+name: "Pull Request Labeler"
+
+on:
+  pull_request_target:
+    types: ["opened"]
+    branches: ["master", "main"]
+
+jobs:
+  labeler:
+    if: github.event.pull_request.user.login == 'devex-sa'
+    permissions:
+      contents: read
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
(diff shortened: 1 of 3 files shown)


.github/workflows/labeler.yml
```diff
@@ -0,0 +1,63 @@
+name: "Pull Request Labeler"
+
+on:
+  pull_request_target:
+    types: ["opened"]
+    branches: ["master", "main"]
+
+jobs:
+  labeler:
+    if: github.event.pull_request.user.login == 'devex-sa'
```
avnes, line 10: I think this should be read from a org secret to avoid exposing too much information about our environment and users.

avnes (changes requested):

.github/workflows/labeler.yml
Mik-Nord: It's unfortunately defined here:

https://github.com/dfds/infrastructure-modules/blob/a5893dbf673c3e6d97efefff4354ab1da7fd0c2d/renovate.json#L4

And we can't use secrets in a json.

avnes (approved):

PR #1086, opened 2023-08-15
Refactoring org-policies to TF objects.

silviucalin:
As previously tested this SCP, I've updated the Restrictive Policy for pipeline runs.
Please have a look on this.


security/org-policies/policies/RestrictivePolicy.json
```diff
@@ -283,6 +283,69 @@
...
+              "backup:UpdateBackupPlan",
+              "backup:UpdateFramework",
+              "backup:UpdateRecoveryPointLifecycle",
+              "backup:UpdateReportPlan"
+            ],
+            "Resource": "*",
+            "Condition": {
+              "StringEquals": {
+                "aws:ResourceTag/dfds.owner": [
+                  "itbuildsourcedevex@dfds.com"
```
avnes, line 318: Keeping an email address in a git repo is sub-optimal, but since this is a JSON file we would need some refactoring first. 

I will approve this PR as is, but would you be so kind to create an issue on the Ops board about refactoring SCP from JSON files to terraform objects?

avnes (dismissed):

security/org-policies/policies/RestrictivePolicy.json
silviucalin: Thanks for the heads up @avnes 
Will try to replace the value with a variable instead rather than merge the PR. Pushing another commit once it's done.

avnes (approved):
Great work!

PR #1286, opened 2024-02-06
This is to allow accounts to be deleted using root on capability deletion process.

silviucalin:
## Describe your changes

During a capability deletion process I identified that a policy change was needed in order to be able to delete the AWS account using the root account.

## Is it a new release?

### DIFF
--- a/security/org-policies/restrictive.tf
+++ b/security/org-policies/restrictive.tf
@@ -136,7 +136,6 @@ data "aws_iam_policy_document" "restrictive" {
     effect = "Deny"
     not_actions = [
       "access-analyzer:*",
-      "account:CloseAccount",
       "account:Get*",
       "account:List*",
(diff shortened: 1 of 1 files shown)


rifisdfds:
I think we may also need to add an explicit deny further down before merging this one

rifisdfds:
I have a feeling that the explicit deny will override the allow, meaning root user is still blocked. I could be wrong but we should test this to be absolutely sure

rifisdfds:
@silviucalin Reading again my messages are slightly confusing. What I mean is that I think it can be allowed by being in the not_action part like you had in your first commit, that bit was fine, and then the deny from your second commit should be the statement that has the condition on it to not include root user of any account, so no need for the allow on the second commit.

That way the deny takes precedence for everyone other than root user

Mik-Nord (approved):

PR #1343, opened 2024-03-25
Added vars for adjusting Grafana agent memory limit and request

SEQUOIIA:


silviucalin (approved):

compute/k8s-services/vars.tf
```diff
@@ -1256,6 +1256,18 @@ variable "grafana_agent_traces_enabled" {
...
 
+variable "grafana_agent_resource_memory_limit" {
+  type = string
+  default = "20Gi"
+  description = "Set resource memory limits on Grafana Agent container"
+}
+
+variable "grafana_agent_resource_memory_request" {
+  type = string
+  default = "4Gi"
```
avnes, line 1267: I think we ought to set this to the same value as max, or at least closer to limit fragmentation.

avnes (approved):

compute/k8s-services/vars.tf
SEQUOIIA: We can do that, but without a dedicated node group it might be difficult to find an available node with that high a request.

PR #1472, opened 2024-07-04
Implement nvidia device plugin flux

Mik-Nord:
## Describe your changes
Removed the old nvidia-device-plugin implementation and added it as a part of flux in the service phase.


compute/k8s-services/vars.tf
```diff
@@ -1493,3 +1493,46 @@ variable "external_secrets_ssm_aws_region" {
...
+variable "deploy_nvidia_device_plugin" {
+  type        = bool
+  default     = false
+  description = "Whether to deploy NVIDIA device plugin. This needs to be set to `true` when GPU based workloads needs to be enabled."
+}
+
+variable "nvidia_chart_version" {
+  type        = string
+  description = "Nvidia device plugin helm chart version"
+  default     = null
```
avnes, line 1510: This must be an empty string and not null.

avnes (changes requested):

avnes (approved):

PR #1528, opened 2024-09-02
feat/alternate contacts

SEQUOIIA:
Allows for setting alternate contacts.


silviucalin (approved):
Let's get this going!

PR #1994, opened 2025-09-16
Fetch dockerhub secrets on the fly, make variables required and move implementaion to containerd config.toml

wcarlsen:
## Describe your changes

* Before we had docker hub credentials in clear text in the Launch Template (not the biggest of problems), but when moving to Karpenter it would be in clear text on a Kubernetes custom resource `EC2NodeClass` in the `spec.userdata` section. Hence moving to fetching them from parameter store on the fly makes sense.
* Docker hub credentials are no longer optional with defaults, so we make sure they are in place once a cluster is spun up. This should add consistency across all clusters from prod -> sandboxes.
* Move implementation from `CredentialsProviderConfig` to `containerd` removes dependencies such as `static-credential-provider` and us managing the `matchImages` from `ecr-credential-provider (also not a huge problem, but it can drift).
* Removed variable `essentials_url` because it is no longer needed.

## Is it a new release?


wcarlsen:
> Damn I have by mistake tested this on an older AMI for some reason. Probably something I inherited from the sandbox-template repo

Okay old AMI seems to not be at fault why the nodes arent coming up. I wonder what it might be then?

wcarlsen:
Okay I found the issue, still doesn't know why though. It seems that on my sandbox cluster this line `[plugins."io.containerd.grpc.v1.cri".registry.configs."registry-1.docker.io".auth]` is okay and works, but for qa `[plugins."io.containerd.grpc.v1.cri".registry.configs."docker.io".auth]` this line works.

wcarlsen:
`registry-1.docker.io` is the cannonical registry host for pulling images for Docker Hub.

SEQUOIIA (approved):
Tested in sandbox cluster, works and pulls private images from Docker Hub. ECR pull from private repo still works.

I personally still prefer the flexibility of the other approach, but I don't care strongly either way.

wcarlsen:
> I personally still prefer the flexibility of the other approach, but I don't care strongly either way.

Just to be clear the flexibility in previous setup caused us not detect that docker hub credentials was filled in the QA pipeline

PR #2025, opened 2025-10-02
Feature/simplification-part-1

avnes:
- [ ] I have rebased the code to master (or merged in the latest from master)

## Is it a new release?
- [ ] Apply a release tag `release:(major|minor|patch)`, following semantic versioning in [this guide](https://semver.org/) or `norelease` if there is no changes to the Terraform code


avnes:
> I think I'm going to need some convincing about why hardcoding a namespace name is better than having it as a variable with a default set that you don't have to touch unless you need to
> 
> This probably needs some discussion

1. I have seen Flux fail if trying to patch a namespace through Terraform. The HelmRelease in platform-apps has a hardcoded namespace anyway, so flux will not be able to patch it when yoy try to patch it in a new namespace.
2. We have just one single use case where we override the namespace, and that is for Traefik where we have a green and blue namespace. Because of the issue in 1) we hence have two Treafik folders in platform-apps with the blue and green namespaces hardcoded.

PR #2038, opened 2025-10-07
EKS nodes version <= 1.34, uses containerd v2 with different syntax

wcarlsen:
## Describe your changes

Prepare for EKS AMI version 1.34 with containerd v2, that uses different syntax see https://github.com/containerd/containerd/blob/main/docs/cri/registry.md.

- [ ] I have rebased the code to master (or merged in the latest from master)

## Is it a new release?
- [ ] Apply a release tag `release:(major|minor|patch)`, following semantic versioning in [this guide](https://semver.org/) or `norelease` if there is no changes to the Terraform code

### DIFF
--- a/_sub/compute/eks-nodegroup-managed/dependencies.tf
+++ b/_sub/compute/eks-nodegroup-managed/dependencies.tf
@@ -39,3 +39,7 @@ locals {
 data "aws_eks_cluster" "this" {
   name = var.cluster_name
 }
+
+locals {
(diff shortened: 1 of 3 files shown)


wcarlsen:
This logic seems to not be correct, we might have to consider that this is not good enough see issues https://github.com/awslabs/amazon-eks-ami/issues/2470#issue-3514989135 with a timeline of when containerd v2 is backported.

wcarlsen:
One could also consider if it by now was worth having the userdata script check if containerd v2 is present?

PR #2342, opened 2026-05-18
enable loading grafana credentials using 1password

samidbb:
## Describe your changes

This PR enables Grafana agent to load secrets from 1password

----
**Detailed description of what has changed:**

**Secrets Management and IRSA Integration:**
* Added a new SSM parameter for storing the 1Password Connect token specifically for Grafana, and exposed its ARN as an output (`_sub/security/helm-1password-connect/main.tf`, `_sub/security/helm-1password-connect/outputs.tf`, `_sub/security/helm-1password-connect/vars.tf`). [[1]](diffhunk://#diff-7ddbd02e75b84f6c3c7481be027e66cd42904eeb3e3ee6916b92995370bd5751R35-R42) [[2]](diffhunk://#diff-4161b0e8cb0e831427c8f150c27466e3dd502f917dc84f958590b22d50463ba1R1-R4) [[3]](diffhunk://#diff-6ce25354b98d018d52fc5c85fabe684cfd758d22ce9c60e924c94ce638d2aa19R72-R78)
* Introduced an IRSA module to create IAM roles for Grafana, allowing it to securely access the SSM parameter containing the 1Password token (`_sub/monitoring/grafana/iam.tf`).
* Added necessary AWS data sources to support IRSA setup (`_sub/monitoring/grafana/dependencies.tf`).

**Configuration and Variable Updates:**
* Passed new variables (`grafana_stack`, `onepassword_access_parameter_store_arn`) through module and variable definitions to enable correct wiring of configuration and secrets (`compute/k8s-services/main.tf`, `compute/k8s-services/vars.tf`, `_sub/monitoring/grafana/vars.tf`). [[1]](diffhunk://#diff-9dcb2b34c331c83c7b15b8785fc38080f22e765d3119f854acc7848e7f669485L620-R621) [[2]](diffhunk://#diff-83380a112934c41699d826a5e1a07c6ee2d0729366db765c2d9c40fd2967c45bR473-R478) [[3]](diffhunk://#diff-f1734db07cbf6166c8f2c16ebd549e7ac676848b244b01dd1e6f28d011986d94R158-R167)
* Updated the GitHub Actions workflow to include the new 1Password token for Grafana (`.github/workflows/qa.yml`). [[1]](diffhunk://#diff-4d515b7cca4aa254976e10e863f94e4527f5edd15f4315279f578385fc0f68b8R112) [[2]](diffhunk://#diff-4d515b7cca4aa254976e10e863f94e4527f5edd15f4315279f578385fc0f68b8R194)

**Helm and Flux Integration:**
* Updated Helm and Flux configuration templates to substitute in the new variables and remove hardcoded secrets from values files (`_sub/monitoring/grafana/main.tf`, `_sub/monitoring/grafana/values/app-config.yaml`, `_sub/monitoring/grafana/values/patch.yaml`). [[1]](diffhunk://#diff-df814ceef35f1c9f7b4ed21fce179164289587b8468bec70ad5c7992ed4b996cR10-R12) [[2]](diffhunk://#diff-457f8d98341bd99a6a3fb3f6362e38cf7925704c64d2c91d2975b859de9c9443R28-R32) [[3]](diffhunk://#diff-6eb8223c7186b318730b8f4ef7b0230bf032e270f0467b41b45ad08fd9bcef52L8-L9) [[4]](diffhunk://#diff-6eb8223c7186b318730b8f4ef7b0230bf032e270f0467b41b45ad08fd9bcef52L64-L92)

## Is it a new release?


wcarlsen:
I do not know our setup for 1password tokens and 1password connect that well, but it seems kinda rough that we have at least 3 1password tokens (most likely not the same), but I suspect they provide the same functionality in the end.

```
TF_VAR_onepassword_token_for_grafana: op://CloudEng-General/1Password Connect/nonprod_grafana_token # pragma: allowlist-secret
TF_VAR_onepassword_token_for_atlantis: op://CloudEng-General/1Password Connect/nonprod_atlantis_token # pragma: allowlist-secret
TF_VAR_onepassword_token_for_grafana: op://CloudEng-General/1Password Connect/nonprod_grafana_token # pragma: allowlist-secret
```

Could this in reality just be one token?

_sub/monitoring/grafana/iam.tf
```diff
@@ -0,0 +1,56 @@
...
+
+    condition {
+      test     = "StringEquals"
+      values   = ["system:serviceaccount:grafana:${local.deploy_name}"]
+      variable = "${local.oidc_issuer}:sub"
+    }
+  }
+}
+
+resource "aws_iam_role" "this" {
```
wcarlsen: Personally I would prefer to use this community module for IRSA roles, but it is a minor thing https://github.com/terraform-aws-modules/terraform-aws-iam/blob/master/modules/iam-role-for-service-accounts/README.md.

wcarlsen (changes requested):
It would seem to me that I can deploy Grafana with `grafana_deploy=true` variable, but provide no 1password token `token_for_grafana` variable, because default is empty string. Would it work then?

I think we need to make sure that if we deploy Grafana then token is also provided, since it seems kinda crucial here. FYI variable validation cannot depend on other variables.

Also see this comment about multiple 1password tokens https://github.com/dfds/infrastructure-modules/pull/2342#issuecomment-4490425605.

wcarlsen:
> I do not know our setup for 1password tokens and 1password connect that well, but it seems kinda rough that we have at least 3 1password tokens (most likely not the same), but I suspect they provide the same functionality in the end.
> 
> ```
> TF_VAR_onepassword_token_for_grafana: op://CloudEng-General/1Password Connect/nonprod_grafana_token # pragma: allowlist-secret
> TF_VAR_onepassword_token_for_atlantis: op://CloudEng-General/1Password Connect/nonprod_atlantis_token # pragma: allowlist-secret
> TF_VAR_onepassword_token_for_grafana: op://CloudEng-General/1Password Connect/nonprod_grafana_token # pragma: allowlist-secret
> ```
> 
> Could this in reality just be one token?

You don't have to solve this issue in this PR, unless you want to, since I will create a PBI for only providing one token for enabling 1password connect access.

.github/workflows/qa.yml
```diff
@@ -190,6 +191,7 @@ jobs:
           TF_VAR_grafana_agent_tempo_username: op://CloudEng-General/Grafana.com - sandbox/Tempo_Username # pragma: allowlist-secret
           TF_VAR_onepassword_credentials_json: op://CloudEng-General/1Password Connect/nonprod_json # pragma: allowlist-secret
           TF_VAR_onepassword_token_for_atlantis: op://CloudEng-General/1Password Connect/nonprod_atlantis_token # pragma: allowlist-secret
+          TF_VAR_onepassword_token_for_grafana: op://CloudEng-General/1Password Connect/nonprod_grafana_token # pragma: allowlist-secret
```
wcarlsen, line 194: this is here twice

wcarlsen (changes requested):

_sub/monitoring/grafana/iam.tf
samidbb: The code is now updated to use the tf module that is referred

.github/workflows/qa.yml
wcarlsen: This is still an issue.

compute/k8s-services/main.tf
```diff
@@ -594,30 +594,31 @@ module "elb_inactivity_cleanup_auth" {
...
-  open_cost_enabled             = var.grafana_agent_open_cost_enabled
-  agent_resource_memory_limit   = var.grafana_agent_resource_memory_limit
-  agent_resource_memory_request = var.grafana_agent_resource_memory_request
-  affinity                      = var.observability_affinity
-  tolerations                   = var.observability_tolerations
-  agent_replicas                = var.grafana_agent_replicas
-  storage_size                  = var.grafana_agent_storage_size
-
+  source                                 = "../../_sub/monitoring/grafana"
+  count                                  = var.grafana_deploy ? 1 : 0
```
wcarlsen: We should block deployment or throw a warning if the token isn't there, no? Because this could lead to a situation where deploy is true and the token is empty and that wouldn't fly.

compute/k8s-services/vars.tf
```diff
@@ -470,6 +470,12 @@ variable "grafana_agent_storage_size" {
   default     = "5Gi"
 }
 
+variable "grafana_stack" {
+  type        = string
+  description = "The Grafana Cloud stack to use"
+  default     = "sandbox"
```
wcarlsen: I understand why you want a default here, but for production and standby this wouldn't be a sane default.

Can grafana_stack only take a few valid values, like an enum? If that is the case I think we should use the is_sandbox variable thingy and use that to select grafana_stack value.

.github/workflows/qa.yml
samidbb: one for each stage? I followed the atlantis example

wcarlsen (changes requested):
Overall this looks good and you are removing patching and using upstream modules. I still think the issue of no coupling between deploy toggle and the token is an issue that hasn't been addressed.

compute/k8s-services/vars.tf
samidbb: I also considered using "is_sandbox" variable but the problem arises when we run staging which it the strange hybrid animal in this chain

wcarlsen: Which stack is staging pointing to? Staging is considered sandbox https://github.com/dfds/eks-pipeline/blob/8da1ac8ac7dae0b86b24f8a194adf5e959d93e0f/staging-oxygen-account/eu-west-1/k8s-staging-hellman/cluster/env.hcl#L16

compute/k8s-services/main.tf
samidbb: I have made a change that should block the deployment on missing token

compute/k8s-services/vars.tf
samidbb: maybe I was thinking about something else. but if we go agree that all non-prod environment cluster are sandbox then it would make it easier to deal with

samidbb: now it is removed in favor of local variable

.github/workflows/qa.yml
wcarlsen: Ahh sorry I assumed that we used .1password files. I got confused sorry.

wcarlsen: We should really use the OP_ENV_FILE feature here.

compute/k8s-services/vars.tf
```diff
@@ -575,6 +575,15 @@ variable "onepassword_token_for_atlantis" {
   description = "The 1Password Connect tokens to be stored in SSM if Atlantis is enabled"
 }
 
+variable "onepassword_token_for_grafana" {
+  type        = string
+  sensitive   = true
+  default     = ""
+  description = <<-EOT
```
wcarlsen, line 585: I really like this description. Could we add the same to Grafana depoy variable for documentation purpose?

wcarlsen (changes requested):
One small thing. Everything else looks good to me.

wcarlsen:
The failure is on an unrelated gateway test. Let's rerun and see if it gets fixed?

compute/k8s-services/vars.tf
samidbb: done

wcarlsen (approved):
Great

PR #2358, opened 2026-05-27
revert deletion: add S3 replication destination module with necessary configurations and variables

avnes:
revert deletion: add S3 replication destination module back


samidbb:
is there a way to have a test of this module here since it is part of eks pipeline?

samidbb (approved):
it would be great to have a test of this module since it is part of eks pipeline

PR #2370, opened 2026-06-02
Add/alb/targetgroups arn output

wcarlsen:
## Describe your changes

In our current setup all nodes are target for the load balancers in place exposing two nodeports each. The targetgroup registration is happening through an autoscalinggroup attachment. This makes all nodes go through a registration and deregistration process and adding on top of this all requests are very, very likely to make an extra network jump from the node to a node that actually has Traefik running.

This PR attempt to do a couple of things:
* Remove autoscalinggroup attachment for Traefik green variant
* Output and use targetgroup arns in TargetgroupBinding via postbuildSubstitution for AWS Loadbalancer Controller to register and deregister targets, in this case Traefik green variant pods
* Change target group type from instance to IP aware so we do not need nodeports exposed on nodes
* On platform-apps side of things we implement an namespace label for AWS Loadbalancer Controller to inject a readiness gate into our Traefik green pods, that will only set a Ready state condition if the pod is registered as a target. This allows for fully graceful deployment rollout of Traefik green without dropping a single request.

See PR for platform-apps https://github.com/dfds/platform-apps/pull/582/changes.

## Is it a new release?


rifisdfds:
I like this a lot, much better than the way we currently do things

only potential issue i can see is the ALB still sending traffic over while it is deregistering a target (we have a 300s delay), but i can't fully get my head around it on a friday. maybe we can try a node rollover with it to see? traefik's prestop option could work around it to keep the pod up i think

wcarlsen:
> I like this a lot, much better than the way we currently do things
> 
> 
> 
> only potential issue i can see is the ALB still sending traffic over while it is deregistering a target (we have a 300s delay), but i can't fully get my head around it on a friday. maybe we can try a node rollover with it to see? traefik's prestop option could work around it to keep the pod up i think

Very good point and we probably should be implementing preStop hooks anyway or similar functionality. I think the Helm Chart for some reason doesn't support it, but that is implementation specific.

wcarlsen:
> > I like this a lot, much better than the way we currently do things
> > only potential issue i can see is the ALB still sending traffic over while it is deregistering a target (we have a 300s delay), but i can't fully get my head around it on a friday. maybe we can try a node rollover with it to see? traefik's prestop option could work around it to keep the pod up i think
> 
> Very good point and we probably should be implementing preStop hooks anyway or similar functionality. I think the Helm Chart for some reason doesn't support it, but that is implementation specific.

The last portion was a lie it is supported https://github.com/traefik/traefik-helm-chart/blob/5bd1ae0c53be8909f4e04332d1c7d79750c72505/traefik/values.yaml#L101.

wcarlsen:
I've implemented preStop hooks in this seperate PR, because I think we need to do it independent of this PR https://github.com/dfds/platform-apps/pull/597.

rifisdfds (approved):

PR #2392, opened 2026-06-23
Migrate traefik blue to IP aware targetgroup

wcarlsen:
## Describe your changes

In our current setup all nodes are target for the load balancers in place exposing nodeports. The targetgroup registration is happening through an autoscalinggroup attachment. This makes all nodes go through a registration and deregistration process and adding on top of this all requests are very, very likely to make an extra network jump from the node to a node that actually has Traefik running.

This PR attempt to do a couple of things:
- Remove autoscalinggroup attachment for Traefik blue variant
- Output and use targetgroup arns in TargetgroupBinding via postbuildSubstitution for AWS Loadbalancer Controller to register and deregister targets, in this case Traefik blue variant pods
- Change target group type from instance to IP aware so we do not need nodeports exposed on nodes
- Removed all nodeports variables related to traefik
- Added moved blocks to indicate resource name change
- On platform-apps side of things we implement an namespace label for AWS Loadbalancer Controller to inject a readiness gate into our Traefik blue pods, that will only set a Ready state condition if the pod is registered as a target. This allows for fully graceful deployment rollout of Traefik blue without dropping a single request.

See PR for platform-apps https://github.com/dfds/platform-apps/pull/609.

I have switched the load balancer weight to blue and added the branch for the PR above, to prove that it works in testing.

To clarify this is the PR that merged for Traefik green https://github.com/dfds/infrastructure-modules/pull/2370. We are running 100% ingress traffic on Traefik green in staging, production and standby.

## Is it a new release?

### DIFF
--- a/test/integration/eu-west-1/k8s-qa/services/terragrunt.hcl
+++ b/test/integration/eu-west-1/k8s-qa/services/terragrunt.hcl
@@ -44,8 +44,8 @@ inputs = {
   # Traefik v2
   # --------------------------------------------------
 
-  traefik_blue_variant_weight   = 0
-  traefik_green_variant_weight  = 1
+  traefik_blue_variant_weight   = 1
+  traefik_green_variant_weight  = 0
 
   # --------------------------------------------------
   # Blaster
@@ -66,7 +66,7 @@ inputs = {
   # Flux CD
   # --------------------------------------------------
 
-  fluxcd_apps_repo_branch           = "main"
+  fluxcd_apps_repo_branch           = "migrate/traefik/blue/ip-aware-targetgroup"
   fluxcd_bootstrap_repo_name        = "platform-manifests-qa"
   fluxcd_bootstrap_repo_branch      = "main"
   fluxcd_version                    = "v2.7.5"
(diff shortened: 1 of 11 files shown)


test/integration/eu-west-1/k8s-qa/services/terragrunt.hcl
```diff
@@ -66,7 +66,7 @@ inputs = {
   # Flux CD
   # --------------------------------------------------
 
-  fluxcd_apps_repo_branch           = "main"
+  fluxcd_apps_repo_branch           = "migrate/traefik/blue/ip-aware-targetgroup"
```
samidbb, line 69: I think will be set back to main when merging the PR?

wcarlsen: Yes I will! This is just to prove within our testing framework that it works. Since we have tests that checks validates ingress traffic. I will also revert the traffic split back to green, since it reflects staging, production and standby better.

samidbb (approved):
lgtm

PR #2396, opened 2026-07-02
Always provision source controller IRSA role with the cluster and allow for pulls through pull through cache

wcarlsen:
## Describe your changes

Currently source controller IRSA role is a shared role from shared-prod account, but since we implement allowed pulls if it is form an org account, we no longer need the shared account approach. So we let the clusters bring its own. We also solve a bug with the shared-prod IRSA role missing permissions to pull images from the pull through cache. Lastly this hopefully removes flaky diffs and random failures on fluxcd_bootstrap_git resource as described here https://dfds.slack.com/archives/CQD754GGL/p1782994535687589.

## Is it a new release?


wcarlsen:
I guess that triggered a bug in the provider 🙈

wcarlsen:
> I guess that triggered a bug in the provider 🙈

So on first run this fails, with an inconsistent plan from the provider, but it disappears on second run.

wcarlsen:
Added bug and blocked labels because there is a bug in the provider.

avnes (dismissed):
LGTM. During my testings I did not experience any issues with this. But I was using the latest and greatest provider versions in my sandbox.

```
provider "registry.opentofu.org/fluxcd/flux" {
  version     = "1.9.3"

provider "registry.opentofu.org/hashicorp/tls" {
  version     = "4.3.0"

provider "registry.opentofu.org/integrations/github" {
  version     = "6.13.0"
```

